PR #8, opened 2022-01-03
Capitalize the first letter of the docstring

### LINKED ISSUE #3: Add capitalization of first letter check
This should also consider argument names for function and class definitions so that they can be excluded from this rule.

Pierre-Sassoulas:
This closes #3

### DIFF
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -16,18 +16,22 @@ class BeginningQuotesFormatter(StringFormatter):
         return new_string
 
 
-class CapitalizeFirstLetter(StringFormatter):
+class CapitalizeFirstLetterFormatter(StringFormatter):
     """Capitalize the first letter of the docstring if appropriate."""
 
     name = "capitalize-first-letter"
-    SNAKE_CASE = re.compile(r"([^\W\dA-Z]+_[^\WA-Z]*)$")
+    first_letter_re = re.compile(r"""['"]{3}\s*(\w)""")
 
     def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
-        new = tokeninfo.string[3:].lstrip()
-        first_space_index = new.find(" ")
-        if self.SNAKE_CASE.match(new[:first_space_index]):
-            return tokeninfo.string
-        return tokeninfo.string[:3] + new[0].upper() + new[1:]
+        new_string = None
+        if match := self.first_letter_re.match(tokeninfo.string):
+            first_letter = match.end() - 1
+            new_string = (
+                tokeninfo.string[:first_letter]
+                + tokeninfo.string[first_letter].upper()
+                + tokeninfo.string[first_letter + 1 :]
+            )
+        return new_string or tokeninfo.string
 
 
 class ClosingQuotesFormatter(StringFormatter):
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -20,9 +20,13 @@ class CapitalizeFirstLetter(StringFormatter):
     """Capitalize the first letter of the docstring if appropriate."""
 
     name = "capitalize-first-letter"
+    SNAKE_CASE = re.compile(r"([^\W\dA-Z]+_[^\WA-Z]*)$")
 
     def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
         new = tokeninfo.string[3:].lstrip()
+        first_space_index = new.find(" ")
+        if self.SNAKE_CASE.match(new[:first_space_index]):
+            return tokeninfo.string
         return tokeninfo.string[:3] + new[0].upper() + new[1:]
 
 
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -16,6 +16,16 @@ class BeginningQuotesFormatter(StringFormatter):
         return new_string
 
 
+class CapitalizeFirstLetter(StringFormatter):
+    """Capitalize the first letter of the docstring if appropriate."""
+
+    name = "capitalize-first-letter"
+
+    def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
+        new = tokeninfo.string[3:].lstrip()
+        return tokeninfo.string[:3] + new[0].upper() + new[1:]
+
+
 class ClosingQuotesFormatter(StringFormatter):
     """Fix the position of the closing quotes."""
 
--- a/tests/data/format/capitalization_first_letter/function_docstrings.py
+++ b/tests/data/format/capitalization_first_letter/function_docstrings.py
@@ -1,0 +1,27 @@
+def func():
+    """
+    a docstring"""
+
+    def inner_func():
+        """
+        a docstring"""
+
+
+def func(named_parameter: str) -> None:
+    """named_parameter should have a nice name
+
+    Other information here
+    """
+    print(named_parameter)
+
+
+def func():
+    """
+    a multi-line
+    docstring
+    """
+
+    def inner_func():
+        """
+        a multi-line
+        docstring"""
--- a/tests/data/format/capitalization_first_letter/function_docstrings.py
+++ b/tests/data/format/capitalization_first_letter/function_docstrings.py
@@ -8,7 +8,7 @@ def func():
 
 
 def func(named_parameter: str) -> None:
-    """named_parameter should have a nice name
+    """'named_parameter' should have a nice name
 
     Other information here
     """
--- a/README.md
+++ b/README.md
@@ -110,6 +110,8 @@ multi-line docstring
 **PEP 257: _The docstring is a phrase ending in a period & Multi-line docstrings consist
 of a summary line just like a one-line docstring_**
 
+Since the first line should be a phrase or summary the first character gets capitalized.
+
 ```python
 # Bad
 """My docstring"""
--- a/pydocstringformatter/formatting/__init__.py
+++ b/pydocstringformatter/formatting/__init__.py
@@ -5,7 +5,7 @@ from typing import List
 from pydocstringformatter.formatting.base import Formatter
 from pydocstringformatter.formatting.formatter import (
     BeginningQuotesFormatter,
-    CapitalizeFirstLetter,
+    CapitalizeFirstLetterFormatter,
     ClosingQuotesFormatter,
     FinalPeriodFormatter,
     SplitSummaryAndDocstringFormatter,
@@ -17,11 +17,9 @@ from pydocstringformatter.formatting.formatter import (
 #   Determine if multi-line or single line and position quotes accordingly
 #   String manipulation in which being multi-line or single line matters
 FORMATTERS: List[Formatter] = [
-    # capitalization needs to be done first because we're stripping
-    # the string to get the first character
-    CapitalizeFirstLetter(),
     SplitSummaryAndDocstringFormatter(),
     BeginningQuotesFormatter(),
     ClosingQuotesFormatter(),
+    CapitalizeFirstLetterFormatter(),
     FinalPeriodFormatter(),
 ]
--- a/pydocstringformatter/formatting/__init__.py
+++ b/pydocstringformatter/formatting/__init__.py
@@ -5,6 +5,7 @@ from typing import List
 from pydocstringformatter.formatting.base import Formatter
 from pydocstringformatter.formatting.formatter import (
     BeginningQuotesFormatter,
+    CapitalizeFirstLetter,
     ClosingQuotesFormatter,
     FinalPeriodFormatter,
     SplitSummaryAndDocstringFormatter,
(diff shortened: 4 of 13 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 pydocstringformatter.egg-info
 dist/
+__pycache__/
```
DanielNoord, line 3: Isn't `pycache` automatically ignored like `.mypy_cache` etc? It's ignored for me.

tests/data/format/capitalization_first_letter/function_docstrings.py
```diff
@@ -0,0 +1,19 @@
+def func():
+    """
+    a docstring"""
+
+    def inner_func():
```
DanielNoord, line 5: Could you add some tests for:
```python
def func(named_parameter: str) -> None:
    """named_parameter should have a nice name
    
    Other information here
    """
    print(named_parameter)
```

Whenever the sentence starts with one of the parameter names I don't think we should enforce capitalisation.

setup.py
```diff
@@ -0,0 +1,3 @@
+from setuptools import setup
+
+setup()
```
DanielNoord, line 3: https://setuptools.pypa.io/en/latest/setuptools.html

I don't think this is necessary from `40.9.0`.

I'd rather include a `[build-system]` section than add an additional file.

.gitignore
Pierre-Sassoulas: I can add this to my global gitignore :)

tests/data/format/capitalization_first_letter/function_docstrings.py
Pierre-Sassoulas: Did not thougth of that, nice catch. Shouldn't a parameter name be inside ``'`` ? I think we should detect it anyway but maybe the result could be made from another Formatter :
```
def func(named_parameter: str) -> None:
    """'named_parameter' should have a nice name
    
    Other information here
    """
    print(named_parameter)
```  ?

File: setup.py (hunk 3 above)
Comment (Pierre-Sassoulas) [reply setup.py:3]: I did not manage to do an editable install or launch pytest in terminal, I'm going to try an up to date setuptools.

Comment (Pierre-Sassoulas) [reply setup.py:3]:
```
      Successfully uninstalled setuptools-44.0.0
Successfully installed setuptools-60.2.0
```

```
main  pip3 install -e .
ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /home/pierre/pydocstringformatter
(A "pyproject.toml" file was found, but editable mode currently requires a setup.py based build.)
```

File: tests/data/format/capitalization_first_letter/function_docstrings.py (hunk 2 above)
Comment (DanielNoord) [reply tests/data/format/capitalization_first_letter/function_docstrings.py:5]: If you look for `Do not use the Emacs convention of mentioning the arguments of functions or methods in upper case in running text.` in https://www.python.org/dev/peps/pep-0257/ you can find the (I think) only mention of argument names.
I'm on the fence about the `'`. If you want to enforce that a multi-line docstring starts with a single sentence and also enforce a short max line length adding `'` can eat into your precious character count..

Note that PEP 8 says that for docstrings the max length is 72... "For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters." at https://www.python.org/dev/peps/pep-0008/


I'd say we can make enforcing `'` an optional check and for now only focus on non-capitalisation of parameter names.

File: setup.py (hunk 3 above)
Comment (DanielNoord) [reply setup.py:3]: Are you sure your `pip3` installs to PATH? 

```console
pip install -e .
Obtaining file:///Users/daniel/DocumentenLaptop/Programming/Github/DanielNoord/docstringformatter
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Requirement already satisfied: tomli>=1.1.0 in /Users/daniel/.pyenv/versions/3.10.0/envs/docstring-3.10.0/lib/python3.10/site-packages (from pydocstringformatter==0.3.0+dev) (1.2.3)
Installing collected packages: pydocstringformatter
  Attempting uninstall: pydocstringformatter
    Found existing installation: pydocstringformatter 0.2.0+dev
    Uninstalling pydocstringformatter-0.2.0+dev:
      Successfully uninstalled pydocstringformatter-0.2.0+dev
  Running setup.py develop for pydocstringformatter
Successfully installed pydocstringformatter-0.3.0+dev
WARNING: You are using pip version 21.2.3; however, version 21.3.1 is available.
You should consider upgrading via the '/Users/daniel/.pyenv/versions/3.10.0/envs/docstring-3.10.0/bin/python3.10 -m pip install --upgrade pip' command.
```

It works fine for me..

Pierre-Sassoulas: I can launch the test from my IDE, it's mouse heavy but it works. I'll resolve this

DanielNoord: I've just added the `build-system` to `pyproject.toml`. Perhaps it works now?

What does `which setuptools` return?

tests/data/format/capitalization_first_letter/function_docstrings.py
Pierre-Sassoulas: We might have to give the previous token(s?) to the Formatters. We need more than the docstring token if we need to get the name of the function argument. We can add optional argument to ``treat_token`` for all formatters, or initialize the Formatters with optional value for all formatters, or do a specific treatment for ``ContextAwareFormatter``... any ideas welcome :)

DanielNoord: I like `ContextAwareFormatter` as it clearly differentiates different types of formatters. However, I worry how it would like in `_format_file`. Do we need `formatting.FORMATTERS` and `formatting.CONTEXT_AWARE_FORMATTERS` then?

Adding an optional argument to `treat_token` might make more sense, but we need to think about what to pass then:
only one token or a list of tokens? And if we pass a list of tokens, how are we going to now which tokens to include in that list?

DanielNoord: It might be easier not to use `FORMATTERS` but call each formatter in `_format_file` individually. That gives us some more freedom in choosing what to pass to each formatter.

Pierre-Sassoulas: > Do we need formatting.FORMATTERS and formatting.CONTEXT_AWARE_FORMATTERS then?

I was thinking of ``if is_instance(formatter, ContextAwareFormatter)`` but this hurts the liskov substitution princple. It's still less hacky than detecting snake_case with regex like I did in the latest commit though :troll:. Maybe two lists is better.

Pierre-Sassoulas: Alternatively maybe instantiating the tokens in the Formatters is not that costly performance wise and we could store context in formatters all the time ?

DanielNoord: One of the problems I foresee is if we want to provide different `context` to different formatters.
What if one needs the previous tokens until the last `class/def` call while the other wants to see if there is a new line in between the docstring and the `__init__` declaration (the latter is a PEP recommendation)?

DanielNoord: > Alternatively maybe instantiating the tokens in the Formatters is not that costly performance wise and we could store context in formatters all the time ?

Based on my experience with `pylint` I know that iterating over `tokens` can be very costly. Remember that PR about a thousand empty tuples?

I think it makes more sense to ditch the `FORMATTERS` list and just call each formatter independently. If we want to keep the call small we could even combine formatters in a parent class and call that class.

Pierre-Sassoulas: Indeed if we're going to create a special case for each specific token we need to give it's defeating the purpose of the design in the first place. If we give all the tokens all the time it would solves the issue. Do you want me to do some performance check ?

DanielNoord: Not sure if it defeats the purpose of the design. I think the small size of the `BeginningQuotesFormatter` shows how effective it already is.

We can see if passing all tokens works. It's probably easier to pass all tokens now and narrow it down later if we want to then create a design based on subsets of tokens that gradually expands.

Pierre-Sassoulas: > we could even combine formatters in a parent class and call that class.

I like the idea of combining all the formatters that needs token in the same class. So we could have a list of `SIMPLE_FORMATTERS` and a class with all the context aware formatters.

DanielNoord: Btw, just getting ahead of myself: we'll likely also need to pass `self.config` at some point to allow optional formatters to see if they should run.. 😅

Pierre-Sassoulas:
I think we need #15 first so we don't have to rebase it.

tests/data/format/capitalization_first_letter/function_docstrings.py
Pierre-Sassoulas: I have a context aware formatter ready but maybe this hack with regex is enough, let me know :)

DanielNoord: This could be really stupid (and I'm not on regex101 right now), but what about:
```python
def func(param):
    """param is a nice parameter"""
```

This regex doesn't catch that right?

DanielNoord:
I think this should be the last feature to go in `0.3.0` and then we can release.

I have been thinking about a good way to do the `param` finding. What about seeing of the previous token is `:` and then iterating over all tokens in reverse order until you find a `def` or `class`? For `module` or `variable` docstring I don't think the `param` handling is necessary so I think this might work?
From the tokens between `def`/`class` till `:` we should then be able to deduct the param names either through joining all strings and using a regex or doing some checks for each token, ie. `if next_token.string == ","`.

DanielNoord:
@Pierre-Sassoulas I've been thinking about this a lot and can't come up with a good reliable way to extract the parameters from a function. I considered doing regex and `ast` parsing, but it's just very difficult to come up with a pattern that reliably extracts all and the correct function definitions.

Should we just go ahead and merge this without it?

Pierre-Sassoulas:
> Should we just go ahead and merge this without it?

Yeah we could use astroid to do that but it's overkill, right ? Also it would make the startup soooo slow. 

I think this is enough, there's a possibility that the very first word of the docstring is a one word function parameter but then it's possible to put ``'`` around it to disambiguate.

DanielNoord:
> Yeah we could use astroid to do that but it's overkill, right ? Also it would make the startup soooo slow.


I tried doing it with a regex pattern that captured the docstring + related `def ...` and then let `ast` parse it. That shouldn't be too slow, but getting a pattern that captures all forms of typing, new lines, comma's, indentation etc. for function definitions right proved very difficult (so I gave up 😅 )


> I think this is enough, there's a possibility that the very first word of the docstring is a one word function parameter but then it's possible to put `'` around it to disambiguate.

Agreed.

Pierre-Sassoulas:
Should we just squash merge then ?

DanielNoord:
Well, tests are failing 😅

I also think we no longer need the regex pattern?

Pierre-Sassoulas:
Ha yes. The regex pattern is handling multi-word variables, we could change the test from a one word variable to a two word variables.

pydocstringformatter/formatting/formatter.py
```diff
@@ -16,6 +16,20 @@ def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
         return new_string
 
 
+class CapitalizeFirstLetter(StringFormatter):
+    """Capitalize the first letter of the docstring if appropriate."""
+
+    name = "capitalize-first-letter"
+    SNAKE_CASE = re.compile(r"([^\W\dA-Z]+_[^\WA-Z]*)$")
+
+    def _treat_string(self, tokeninfo: tokenize.TokenInfo) -> str:
```
DanielNoord, line 25:
```suggestion
    def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
```

pydocstringformatter/formatting/formatter.py
```diff
@@ -16,6 +16,20 @@ def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
...
+        new = tokeninfo.string[3:].lstrip()
```
DanielNoord, line 26: Thinking about this: should the `lstrip` be in this formatter? Perhaps not right? This should simply return `tokeninfo.string[:3] + tokeninfo.string[4].upper() + tokeninfo.string[5:]` while we could also add a `DocstringWhitespaceStripperFormatter` (what a name...).

DanielNoord: Added benefit of this would be that this checker wouldn't be order dependent anymore and we can keep the order dependent formatters to a minimum and self-explanatory. ie: it's logical that a whitespace stripper should go first, it is not immediately logical why a capitaliser should go first.

Pierre-Sassoulas: We would need to detect the first letter in case there's multiple spaces, it would impact performance a little. But I agree that doing one thing only is important.

DanielNoord: Hm, what if a user disabled `trim-whitespace-formatter` but does enable `capitalise-first-letter-formatter`. Should we then "capitalise the whitespace" or capitalise the first letter we find?

Pierre-Sassoulas: Capitalizing the whitespace does nothing so I think it's not intuitive and we should capitalize the first letter ?

DanielNoord: Yeah I guess. Can we do that with `r"\['"]+\s*(\w)"`? That should get the first letter after any possible whitespaces and removes the need to strip etc.

DanielNoord:
@Pierre-Sassoulas Changed some things around. This should be good to go now (I think).

Pierre-Sassoulas (commented):
I can't approve my own pull request but this regex is pretty nice, I think we can merge !

PR #9, opened 2022-01-03
Refactor of the formatter to scale better later on

Pierre-Sassoulas:
This is a proposition so it's easier to add Formatter later on and remove some duplicated calls. (Based on #8)

### DIFF
--- a/pydocstringformatter/formatting.py
+++ b/pydocstringformatter/formatting.py
@@ -1,39 +1,74 @@
+import abc
 import re
 import tokenize
 
+# pylint: disable=too-few-public-methods
+from typing import List
 
-def _format_beginning_quotes(tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+
+class Formatter:
+    """Base class for docstring formatter"""
+
+    @abc.abstractmethod
+    def treat(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+        """Return a modified token"""
+
+
+class StringFormatter(Formatter):
+    """Base class for formatter that only modify the string content"""
+
+    @abc.abstractmethod
+    def _treat_string(self, new_string: str) -> str:
+        """Will modify the string"""
+
+    def treat(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+        return tokenize.TokenInfo(
+            tokeninfo.type,
+            self._treat_string(tokeninfo.string),
+            tokeninfo.start,
+            tokeninfo.end,
+            tokeninfo.line,
+        )
+
+
+class BeginningQuotesFormatter(StringFormatter):
     """Fix the position of the opening quotes"""
-    new_string = tokeninfo.string
-    if new_string[3] == "\n":
-        new_string = re.sub(r"\n *", "", new_string, 1)
-    return tokenize.TokenInfo(
-        tokeninfo.type, new_string, tokeninfo.start, tokeninfo.end, tokeninfo.line
-    )
+
+    def _treat_string(self, new_string: str) -> str:
+        if new_string[3] == "\n":
+            new_string = re.sub(r"\n *", "", new_string, 1)
+        return new_string
 
 
-def _format_multiline_ending_quotes(
-    tokeninfo: tokenize.TokenInfo,
-) -> tokenize.TokenInfo:
-    """Fix the position of end quotes for multi-line docstrings"""
-    good_end = f"{(tokeninfo.start[1]) * ' '}{(tokeninfo.string[0]) * 3}"
-    new_string = tokeninfo.string
-    split_string = new_string.split("\n")
-
-    # Add new line with only quotes
-    if not new_string.endswith("\n" + good_end):
-        new_string = new_string[:-3] + "\n" + good_end
-    # Remove line with only quotes for potential single line string
-    elif len(split_string) == 2 and split_string[-1] == good_end:
-        new_string = "\n".join(split_string[:-1]) + tokeninfo.string[0] * 3
-
-    return tokenize.TokenInfo(
-        tokeninfo.type, new_string, tokeninfo.start, tokeninfo.end, tokeninfo.line
-    )
-
-
-def _format_closing_quotes(tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+class ClosingQuotesFormatter(Formatter):
     """Fix the position of the closing quotes"""
-    if "\n" in tokeninfo.string:
-        return _format_multiline_ending_quotes(tokeninfo)
-    return tokeninfo
+
+    @staticmethod
+    def _format_multiline_ending_quotes(
+        tokeninfo: tokenize.TokenInfo,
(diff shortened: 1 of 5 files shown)


DanielNoord:
I like this approach!

pydocstringformatter/formatting.py
```diff
@@ -1,39 +1,74 @@
+import abc
 import re
 import tokenize
 
+# pylint: disable=too-few-public-methods
```
DanielNoord, line 5: Could you add this to `pyproject.toml`? I keep ignoring this in my projects anyway 😄

pydocstringformatter/formatting.py
```diff
@@ -1,39 +1,74 @@
...
+from typing import List
 
-def _format_beginning_quotes(tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+
+class Formatter:
+    """Base class for docstring formatter"""
+
+    @abc.abstractmethod
+    def treat(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
```
DanielNoord, line 13:
```suggestion
    def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
```

pydocstringformatter/formatting.py
```diff
@@ -1,39 +1,74 @@
...
+        """Return a modified token"""
+
+
+class StringFormatter(Formatter):
+    """Base class for formatter that only modify the string content"""
```
DanielNoord, line 18:
```suggestion
    """Base class for formatter that only modifies the string content"""
```

pydocstringformatter/formatting.py
```diff
@@ -1,39 +1,74 @@
...
+
+    @abc.abstractmethod
+    def _treat_string(self, new_string: str) -> str:
+        """Will modify the string"""
+
+    def treat(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
```
DanielNoord, line 24:
```suggestion
    def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
```

pydocstringformatter/formatting.py
```diff
@@ -1,39 +1,74 @@
...
-
-def _format_closing_quotes(tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+
+    def _treat_string(self, new_string: str) -> str:
+        if new_string[3] == "\n":
+            new_string = re.sub(r"\n *", "", new_string, 1)
+        return new_string
+
+
+class ClosingQuotesFormatter(Formatter):
```
DanielNoord, line 43: Shouldn't this be a `StringFormatter` as well?

pydocstringformatter/formatting.py
Pierre-Sassoulas: Done

DanielNoord:
@Pierre-Sassoulas Just something I found while reading the PEPs (and to be added at a later point) one-liner's and summary lines should be sentences ending with a point. I have not done that myself for some of the docstrings I already wrote, but we might as well start doing this ourselves now.

pydocstringformatter/formatting.py
Pierre-Sassoulas: Done (We should make this one optional asap, it's very annoying to remove it first thing in all new projects 😄)

pydocstringformatter/formatting.py
Pierre-Sassoulas: ``return self._format_multiline_ending_quotes(tokeninfo)`` returns a token

DanielNoord: It does, but only ever changes the `string`. It's a bad design that gets shown by this wonderful refactor! 😄

Pierre-Sassoulas:
I can remove the ``setup.py`` commit so we merge this one first if this is controversial 😉 .

DanielNoord:
> I can remove the `setup.py` commit so we merge this one first if this is controversial 😉 .

Would be good I think! I hate having to scroll down two pages on a Github repository because of all the config files, that's why I'd like to enforce not having a `setup.py`.

I should add a `build-system` section though, so that's on the TODO-list.

pydocstringformatter/formatting.py
Pierre-Sassoulas: Hmm but it needs the token and not the string to work. Maybe treat_string needs to take a token as input ?

Pierre-Sassoulas: Or we could initialize the formatters for each loop and the token become a class attribute.

DanielNoord: Ah that makes sense I think.

DanielNoord: > Or we could initialize the formatters for each loop and the token become a class attribute.

That sounds like a large impact on performance. Would probably still be manageable, but I like how quick the tool runs at the moment.

Pierre-Sassoulas: I've refactored treat_string, the result does look cleaner.

pydocstringformatter/formatting/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+
+from typing import List
+
+from pydocstringformatter.formatting.formatter import (
+    BeginningQuotesFormatter,
+    ClosingQuotesFormatter,
+    Formatter,
+)
+
+FORMATTERS: List[Formatter] = [
```
DanielNoord, line 11:
```suggestion
FORMATTERS = [
```

pydocstringformatter/formatting/__init__.py
```diff
@@ -0,0 +1,14 @@
+__all__ = ["FORMATTERS"]
+
+from typing import List
```
DanielNoord, line 3:
```suggestion
```

pydocstringformatter/formatting/formatter.py
```diff
@@ -0,0 +1,59 @@
...
+    def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+        """Return a modified token"""
+
+
+class StringFormatter(Formatter):
+    """Base class for formatter that only modifies the string content"""
+
+    @abc.abstractmethod
+    def _treat_string(self, tokeninfo: tokenize.TokenInfo) -> str:
+        """Will modifies the string"""
```
DanielNoord, line 19:
```suggestion
        """Return a modified string"""
```

DanielNoord (approved):
Three final comments! I think this can be squashed and merged 😄 

Congratulations on becoming a `pydocstringformatter` contributor 😉

Pierre-Sassoulas:
> Congratulations on becoming a pydocstringformatter contributor

It's actually nice to be on a project with a bug to feature ratio of 0%, how do you do that ? 😄

DanielNoord:
> > Congratulations on becoming a pydocstringformatter contributor
> 
> It's actually nice to be on a project with a bug to feature ratio of 0%, how do you do that ? 😄

Rigorous reviews of all PRs 😄

PR #14, opened 2022-01-03
Add line wrapping to summaries of docstrings

### LINKED ISSUE #7: Cut lines that exceed a number of caracter and reorganize docstrings
Like prettier does. Must not break numpy/google style docstrings. Should not break the first line, except if there is multiple sentence then it should take only the first one and move the rest in the body of the docstring.

Default to 88 char like black, but should probably be configurable ?

Pierre-Sassoulas:
Partially handles #7, it's not even close right now 😄 The line wrapping fail because there are space before and after the leading string.

- [ ] Line wrapping after certain line length
- [ ] make line length configurable, search for settings of other tools (black, pylint)
- [ ] Break up multiple sentences on first line into summary and main block (see PEP 256)
- [ ] Handle numpy/google style docstrings gracefully with line wrapping

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -11,6 +11,7 @@ Current usage of ``pydocstringformatter``:
                                 [--summary-quotes-same-line]
                                 [--split-summary-body  --no-split-summary-body]
                                 [--strip-whitespaces  --no-strip-whitespaces]
+                                [--linewrap-full-docstring  --no-linewrap-full-docstring]
                                 [--beginning-quotes  --no-beginning-quotes]
                                 [--closing-quotes  --no-closing-quotes]
                                 [--capitalize-first-letter  --no-capitalize-first-letter]
@@ -79,3 +80,7 @@ Current usage of ``pydocstringformatter``:
                             currently optional as its considered somwehat
                             opinionated and might require major refactoring for
                             existing projects. (default: False)
+      --linewrap-full-docstring, --no-linewrap-full-docstring
+                            Activate or deactivate linewrap-full-docstring:
+                            Linewrap the docstring by the pre-defined line length.
+                            (default: False)
--- a/pydocstringformatter/formatting/__init__.py
+++ b/pydocstringformatter/formatting/__init__.py
@@ -8,6 +8,7 @@ from pydocstringformatter.formatting.formatter import (
     CapitalizeFirstLetterFormatter,
     ClosingQuotesFormatter,
     FinalPeriodFormatter,
+    LineWrapperFormatter,
     QuotesTypeFormatter,
     SplitSummaryAndDocstringFormatter,
     StripWhitespacesFormatter,
@@ -21,6 +22,7 @@ from pydocstringformatter.formatting.formatter import (
 FORMATTERS: List[Formatter] = [
     SplitSummaryAndDocstringFormatter(),
     StripWhitespacesFormatter(),
+    LineWrapperFormatter(),
     BeginningQuotesFormatter(),
     ClosingQuotesFormatter(),
     CapitalizeFirstLetterFormatter(),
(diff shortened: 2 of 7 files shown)


DanielNoord:
I like how it auto-formatted our own test examples 😄

DanielNoord:
@Pierre-Sassoulas I have started working some more on this as it is a nice little job I can do in between other (personal) stuff.

This is still WIP, but in order for this to remain manageable I'd like to focus on wrapping summaries for now. In this PR we should add that basic functionality with the line length as ordered by PEP 8 (which is 72 😮).
Before releasing this we should probably also allow setting the max line length in some way.

One thing I haven't yet fixed is how to handle docstrings that haven't been split. For example:
```python
"""My beautiful docstring
This function will do a number of things:
bla di bla
"""
```
Should be
```python
"""My beautiful docstring

This function will do a number of things:
bla di bla
"""
```

That makes it much easier for this to line wrap. However since the `\n\n` is missing, this current iteration will now wrap this to:
```python
"""My beautiful docstring This function will do a number of things: bla di bla."""
```

Pierre-Sassoulas:
>  in order for this to remain manageable I'd like to focus on wrapping summaries for now

Yes, let's not do too much at once.

> add that basic functionality with the line length as ordered by PEP 8 (which is 72 😮).

I think this is unreasonable and we should use 88 by default (black's default).

>  this current iteration will now wrap this to: """My beautiful docstring This function will do a number of things: bla di bla."""

Is it really problematic ? I mean if we encounter an end of sentence punctuation we could cut easily after it then line wrap. Maybe we should first add end of sentence punctuation if there's a ``\n`` then line wrap ? If there's no ``\n`` at all in the docstring we do not add a ``.``, and we line wrap ?

DanielNoord:
> > add that basic functionality with the line length as ordered by PEP 8 (which is 72 😮).
> 
> I think this is unreasonable and we should use 88 by default (black's default).

Personally I'd like to strictly adhere to the PEP's and make it very clear how to avoid this (awful) default. Something like "we recommend running with `pydocstringformatter --line-length=88`" in the first sections of the README.
Adhering strictly to the PEP's avoids any future discussion about defaults etc, which is something that's likely going to come up with a strict tool like this.
 
> > this current iteration will now wrap this to: """My beautiful docstring This function will do a number of things: bla di bla."""
> 
> Is it really problematic ? I mean if we encounter an end of sentence punctuation we could cut easily after it then line wrap. Maybe we should first add end of sentence punctuation if there's a `\n` then line wrap ? If there's no `\n` at all in the docstring we do not add a `.`, and we line wrap ?

We do cut in `SplitSummaryAndDocstringFormatter`. So `pydocstringformatter  --linewrap-full-docstring --split-summary-body .` works best. However, that formatter is still optional as I'm not fully satisfied with how we determine what is a summary and what is a description. I think it's a good design philosophy not to depend too much on other checkers being activated. That's also why I use `quotes` here instead of `"""`. If people really want to use `'''` or `"` then this formatter won't break that.

We could also think: let's format everything and if the user doesn't want us to they should clearly indicate what is a summary and what isn't. However, I think that's not the most user-friendly approach.

I'll think about this some more and hopefully come up with something elegant.

Pierre-Sassoulas:
> make it very clear how to avoid this (awful) default

Hmm, why put the bad default value in the first place if we think everyone need to change it ? black is pretty successful with a "sane default" and the possibility to change it to the value from pep8 if you're a purist (and 79 is more reasonable than 72 too). Having good default values is pretty important imo.

> avoids any future discussion

If the default is sane it's going make the discussion less likely. I think that if black default was 79, the first thing to do when adopting black would be to discuss between 79, 80, 90, 100, 110, 120 for line length. As it's reasonable, I think discussions happen less often.

Also it would not be a problem if we go check the line length for black/pylint/flake8 ourselves in the configuration files later on.

> We could also think: let's wrap everything and if the user doesn't want us to they should clearly indicate what is a summary and what isn't

Pretty much agree with that, if the docstring is bad we cannot make it good magically. crap in, crap out !

DanielNoord:
@Pierre-Sassoulas Could you review this 😄

Issue #2, opened 2022-01-02
Print out ``diffs`` instead of complete files


DanielNoord:
When not running in ``--write`` mode the program should print a ``diff`` instead of the complete rewritten file.

PR #15, opened 2022-01-04
Make every formatter optional

Pierre-Sassoulas:
Work in progress, this still need more robust tests. This would permit to automatically generate the arguments and to activate or deactivate every parser while still having a default.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -9,9 +9,9 @@ Use `pre-commit install` to install the pre-commit hook for the repository.
 - Implement a Formatter by inheriting from `pydocstringformatter.formatting.Formatter`
 - Add your new formatter to `pydocstringformatter.formatting.FORMATTERS`
 - Write a clear docstring because this will be user-facing: it's what will be seen in
-  the help message for the formatters command line option.
+  the help message for the formatter's command line option.
 - Choose a proper name because this will be user-facing: the name will be used to turn
-  the checker on and off via the command line or config files.
+  the formatter on and off via the command line or config files.
 
 ### Testing
 
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -8,8 +8,10 @@ Use `pre-commit install` to install the pre-commit hook for the repository.
 
 - Implement a Formatter by inheriting from `pydocstringformatter.formatting.Formatter`
 - Add your new formatter to `pydocstringformatter.formatting.FORMATTERS`
-- Choose a proper name because this will be user-facing: the name will be used for
-  options of the CLI.
+- Write a clear docstring because this will be user-facing: it's what will be seen in
+  the help message for the formatters command line option.
+- Choose a proper name because this will be user-facing: the name will be used to turn
+  the checker on and off via the command line or config files.
 
 ### Testing
 
--- a/pydocstringformatter/run.py
+++ b/pydocstringformatter/run.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import List, Union
 
 from pydocstringformatter import __version__, formatting, utils
+from pydocstringformatter.formatting import FORMATTERS
 
 
 class _Run:
@@ -16,6 +17,7 @@ class _Run:
 
     def __init__(self, argv: Union[List[str], None]) -> None:
         self.arg_parser = utils._register_arguments(__version__)
+        utils._register_arguments_formatters(self.arg_parser, FORMATTERS)
         self.config = argparse.Namespace()
 
         if argv := argv or sys.argv[1:]:
@@ -47,9 +49,11 @@ class _Run:
         for index, tokeninfo in enumerate(tokens):
             new_tokeninfo = tokeninfo
 
+            formatter_options = vars(self.config)
             if utils._is_docstring(new_tokeninfo, tokens[index - 1]):
                 for formatter in formatting.FORMATTERS:
-                    new_tokeninfo = formatter.treat_token(new_tokeninfo)
+                    if formatter_options[formatter.name]:
+                        new_tokeninfo = formatter.treat_token(new_tokeninfo)
             changed_tokens.append(new_tokeninfo)
 
             if tokeninfo != new_tokeninfo:
--- a/pydocstringformatter/run.py
+++ b/pydocstringformatter/run.py
@@ -9,7 +9,6 @@ from pathlib import Path
 from typing import List, Union
 
 from pydocstringformatter import __version__, formatting, utils
-from pydocstringformatter.formatting import FORMATTERS
 
 
 class _Run:
@@ -17,13 +16,13 @@ class _Run:
 
     def __init__(self, argv: Union[List[str], None]) -> None:
         self.arg_parser = utils._register_arguments(__version__)
-        utils._register_arguments_formatters(self.arg_parser, FORMATTERS)
+        utils._register_arguments_formatters(self.arg_parser, formatting.FORMATTERS)
         self.config = argparse.Namespace()
 
         if argv := argv or sys.argv[1:]:
-            utils._parse_toml_file(self.arg_parser, self.config)
-            utils._parse_command_line_arguments(self.arg_parser, argv, self.config)
-
+            utils._parse_options(
+                self.arg_parser, self.config, argv, formatting.FORMATTERS
+            )
             self._check_files(self.config.files)
         else:
             self.arg_parser.print_help()
--- a/pydocstringformatter/run.py
+++ b/pydocstringformatter/run.py
@@ -48,10 +48,9 @@ class _Run:
         for index, tokeninfo in enumerate(tokens):
             new_tokeninfo = tokeninfo
 
-            formatter_options = vars(self.config)
             if utils._is_docstring(new_tokeninfo, tokens[index - 1]):
                 for formatter in formatting.FORMATTERS:
-                    if formatter_options[formatter.name]:
+                    if getattr(self.config, formatter.name):
                         new_tokeninfo = formatter.treat_token(new_tokeninfo)
             changed_tokens.append(new_tokeninfo)
 
--- a/pydocstringformatter/utils/argument_parsing.py
+++ b/pydocstringformatter/utils/argument_parsing.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, List, Optional
 
 import tomli
 
+from pydocstringformatter.formatting.base import Formatter
 from pydocstringformatter.utils.exceptions import TomlParsingError, UnrecognizedOption
 
 OPTIONS_TYPES = {"write": "store_true"}
@@ -36,7 +37,26 @@ def _register_arguments(version: str) -> argparse.ArgumentParser:
         version=version,
         help="Show version number and exit",
     )
+    return parser
 
+
+def _register_arguments_formatters(
+    parser: argparse.ArgumentParser, formatters: List[Formatter]
+) -> argparse.ArgumentParser:
+    """Register a list of formatters, so they can all be deactivated or activated."""
+    for formatter in formatters:
+        name = formatter.name
+        help_text = f"ctivate the {name} formatter"
+        parser.add_argument(
+            f"--{name}",
+            action="store_true",
+            dest=name,
+            default=not formatter.optional,
+            help=f"A{help_text} : {formatter.__doc__}",
+        )
+        parser.add_argument(
+            f"--no-{name}", action="store_false", dest=name, help=f"Dea{help_text}"
+        )
     return parser
 
 
--- a/pydocstringformatter/utils/argument_parsing.py
+++ b/pydocstringformatter/utils/argument_parsing.py
@@ -51,7 +51,6 @@ def _register_arguments_formatters(
             f"--{name}",
             action="store_true",
             dest=name,
-            default=not formatter.optional,
             help=f"A{help_text} : {formatter.__doc__}",
         )
         parser.add_argument(
@@ -100,3 +99,37 @@ def _parse_toml_file(
             arguments += _parse_toml_option(key, value)
 
         parser.parse_args(arguments, namespace)
+
+
+def _load_formatters_default_option(
+    parser: argparse.ArgumentParser,
+    namespace: argparse.Namespace,
+    formatters: List[Formatter],
+) -> None:
+    """Load the list of formatters based on their 'optional' attribute."""
+    arguments: List[str] = []
+    for formatter in formatters:
+        if formatter.optional:
+            arguments.append(f"--no-{formatter.name}")
+        elif not formatter.optional:
+            arguments.append(f"--{formatter.name}")
+
+    parser.parse_known_args(arguments, namespace)
+
+
+def _parse_options(
+    parser: argparse.ArgumentParser,
+    namespace: argparse.Namespace,
+    argv: List[str],
+    formatters: List[Formatter],
+) -> None:
+    """Load all default option values.
+
+    The order of parsing is:
+    1. default values, 2. configuration files, 3. command line arguments.
+    """
+    _load_formatters_default_option(parser, namespace, formatters)
+
+    _parse_toml_file(parser, namespace)
+
+    _parse_command_line_arguments(parser, argv, namespace)
--- a/pydocstringformatter/utils/argument_parsing.py
+++ b/pydocstringformatter/utils/argument_parsing.py
@@ -106,7 +106,7 @@ def _load_formatters_default_option(
     namespace: argparse.Namespace,
     formatters: List[Formatter],
 ) -> None:
-    """Load the list of formatters based on their 'optional' attribute."""
+    """Parse the state of the list of formatters based on their 'optional' attribute."""
     arguments: List[str] = []
     for formatter in formatters:
         if formatter.optional:
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -3,7 +3,6 @@ import os
 import sys
 from pathlib import Path
 from typing import List
-from unittest.mock import patch
 
 import pytest
 
@@ -12,11 +11,13 @@ from pydocstringformatter.formatting import FORMATTERS
 
 
 def test_no_arguments(capsys: pytest.CaptureFixture[str]) -> None:
-    """Test that we warn when no arguments are provided"""
+    """Test that we display a help message when no arguments are provided."""
     sys.argv = ["pydocstringformatter"]
     pydocstringformatter.run_docstring_formatter()
     out, err = capsys.readouterr()
     assert out.startswith("usage: pydocstringformatter [-h]")
+
+    # Test that we print help messages for individual formatters as well
     assert "--beginning-quotes" in out
     assert "Activate the beginning-quotes formatter" in out
     assert "--no-beginning-quotes" in out
@@ -128,11 +129,10 @@ def test_optional_formatters(
     capsys: pytest.CaptureFixture[str],
     tmp_path: Path,
 ) -> None:
-    """Test that optional check are activated or not depending on options."""
+    """Test that (optional) formatters are activated or not depending on options."""
     bad_docstring = tmp_path / "bad_docstring.py"
     bad_docstring.write_text(f'"""{"a" * 120}\n{"b" * 120}"""')
-    with patch.object(sys, "argv", ["pydocstringformatter", str(bad_docstring)] + args):
-        pydocstringformatter.run_docstring_formatter()
+    pydocstringformatter.run_docstring_formatter([str(bad_docstring)] + args)
     out, err = capsys.readouterr()
     assert not err
     if should_format:
--- a/tests/test_run.py
+++ b/tests/test_run.py
@@ -2,19 +2,26 @@
 import os
 import sys
 from pathlib import Path
+from typing import List
+from unittest.mock import patch
 
 import pytest
 
 import pydocstringformatter
+from pydocstringformatter.formatting import FORMATTERS
 
 
 def test_no_arguments(capsys: pytest.CaptureFixture[str]) -> None:
     """Test that we warn when no arguments are provided"""
     sys.argv = ["pydocstringformatter"]
     pydocstringformatter.run_docstring_formatter()
-    output = capsys.readouterr()
-    assert output.out.startswith("usage: pydocstringformatter [-h]")
-    assert not output.err
+    out, err = capsys.readouterr()
+    assert out.startswith("usage: pydocstringformatter [-h]")
+    assert "--beginning-quotes" in out
+    assert "Activate the beginning-quotes formatter" in out
+    assert "--no-beginning-quotes" in out
+    assert "Deactivate the beginning-quotes formatter" in out
+    assert not err
 
 
 def test_sys_agv_as_arguments(
@@ -106,3 +113,33 @@ Formatted {expected_second_path} 📖
 """
     )
     assert not output.err
+
+
+@pytest.mark.parametrize(
+    "args,should_format",
+    [
+        [[f"--no-{f.name}" for f in FORMATTERS], False],
+        [[f"--{f.name}" for f in FORMATTERS], True],
+    ],
+)
+def test_optional_formatters(
+    args: List[str],
+    should_format: bool,
+    capsys: pytest.CaptureFixture[str],
+    tmp_path: Path,
+) -> None:
+    """Test that optional check are activated or not depending on options."""
+    bad_docstring = tmp_path / "bad_docstring.py"
+    bad_docstring.write_text(f'"""{"a" * 120}\n{"b" * 120}"""')
+    with patch.object(sys, "argv", ["pydocstringformatter", str(bad_docstring)] + args):
+        pydocstringformatter.run_docstring_formatter()
+    out, err = capsys.readouterr()
+    assert not err
+    if should_format:
+        msg = "Nothing was modified, but all formatters are activated."
+        assert "Nothing to do!" not in out
+        expected = ["---", "@@", "+++"]
+        assert all(e in out for e in expected), msg
+    else:
+        msg = "Something was modified, but all formatter are deactivated."
+        assert "Nothing to do!" in out, msg
--- a/pydocstringformatter/utils/__init__.py
+++ b/pydocstringformatter/utils/__init__.py
@@ -1,6 +1,5 @@
 from pydocstringformatter.utils.argument_parsing import (
-    _parse_command_line_arguments,
-    _parse_toml_file,
+    _parse_options,
     _register_arguments,
     _register_arguments_formatters,
 )
@@ -18,12 +17,11 @@ __all__ = [
     "_find_python_files",
     "_generate_diff",
     "_is_docstring",
-    "_parse_command_line_arguments",
-    "_parse_toml_file",
     "ParsingError",
     "PydocstringFormatterError",
     "_register_arguments",
     "_register_arguments_formatters",
     "TomlParsingError",
+    "_parse_options",
     "_print_to_console",
 ]
--- a/pydocstringformatter/utils/__init__.py
+++ b/pydocstringformatter/utils/__init__.py
@@ -2,6 +2,7 @@ from pydocstringformatter.utils.argument_parsing import (
     _parse_command_line_arguments,
     _parse_toml_file,
     _register_arguments,
+    _register_arguments_formatters,
 )
 from pydocstringformatter.utils.exceptions import (
     ParsingError,
@@ -22,6 +23,7 @@ __all__ = [
     "ParsingError",
     "PydocstringFormatterError",
     "_register_arguments",
+    "_register_arguments_formatters",
     "TomlParsingError",
     "_print_to_console",
 ]


pydocstringformatter/formatting/base.py
```diff
@@ -5,6 +5,19 @@
 class Formatter:
     """Base class for docstring formatter"""
 
+    optional: bool = False
```
DanielNoord, line 8:
```suggestion
    optional = False
```

pydocstringformatter/formatting/base.py
```diff
@@ -5,6 +5,19 @@
...
+
+    def name(self, delimiter: str = "-") -> str:
+        """Return the class name as a list of word."""
```
DanielNoord, line 11:
```suggestion
        """Return the class name as a list of words."""
```

pydocstringformatter/formatting/base.py
```diff
@@ -5,6 +5,19 @@
...
+        result = ""
+        class_name = self.__class__.__name__.replace("Formatter", "")
+        for i, char in enumerate(class_name):
+            if char.upper() == char and i:
+                result += f"{delimiter}{char.lower()}"
+            else:
+                result += char.lower()
+        return result
```
DanielNoord, line 19: Isn't this a bit overcomplicated? How about just adding a `name` attribute to each class, like `optional`. That way we have some more freedom in option names and class names.
It would also allow double names like "--fix-capitalization" and "--f-c" (not that I want to add --f-c, just an example)

pydocstringformatter/formatting/__init__.py
```diff
@@ -1,11 +1,14 @@
...
+from typing import List
+
+from pydocstringformatter.formatting.base import Formatter
 from pydocstringformatter.formatting.formatter import (
     BeginningQuotesFormatter,
     ClosingQuotesFormatter,
 )
 
-FORMATTERS = [
+FORMATTERS: List[Formatter] = [
```
DanielNoord, line 11:
```suggestion
FORMATTERS = [
```

pydocstringformatter/formatting/base.py
DanielNoord: Shall we add a docstring here?

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
...
         version=version,
         help="Show version number and exit",
     )
+    return parser
+
 
+def _register_arguments_formatters(
+    parser: argparse.ArgumentParser, formatters: List[Formatter]
+) -> argparse.ArgumentParser:
+    """Register a list of formatter, so they can all be deactivated or activated."""
```
DanielNoord, line 43:
```suggestion
    """Register a list of formatters, so they can all be deactivated or activated."""
```

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
...
+    for formatter in formatters:
+        help_text = f"ctivate the {formatter.name(delimiter=' ')} formatter"
```
DanielNoord, line 45:
```suggestion
        help_text = f"Activate the {formatter.name(delimiter=' ')} formatter"
```

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
...
+        name = formatter.name()
+        parser.add_argument(
+            f"--{name}",
+            action="store_true",
+            dest=name,
+            default=not formatter.optional,
```
DanielNoord, line 51: `store_true` implies a `True` default value. Perhaps we should determine the action based on `formatter.optional`?

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
...
+            help=f"A{help_text} : {formatter.__doc__}",
```
DanielNoord, line 52: I like this, but this should probably be documented somewhere in the "Contributing guidelines". This is crucial information to know when writing a new formatter.

DanielNoord (changes requested):
There is also an empty file `pydocstringformatter/formatting/argparse_registerer.py`.

I do like this idea though! What about adding an `option_group` in `argparse`? That way they will be separated in the help message and we have some more control over what to display and what not.

pydocstringformatter/formatting/base.py
Pierre-Sassoulas: I see your point, it feel complicated now but it also means we have nothing to do later and class name is consistent with option name. I'll let you choose and I'll do the modification :)

pydocstringformatter/formatting/__init__.py
Pierre-Sassoulas: Right now it's all StringFormatter so mypy infer it as a List of StringFormatter and it's a problem in `_register_arguments_formatters`. We might want to be excplicit for this one.

pydocstringformatter/utils/argument_parsing.py
Pierre-Sassoulas: Not a typo, the help text is f"A{help_text}" or f"Dea{help_text}" later on.

pydocstringformatter/utils/argument_parsing.py
Pierre-Sassoulas: We have both options ``--closing-quotes`` (that store_true) and  ``--no-closing-quotes`` (that store_false). Whatever the default use using one option or the other will change that default value.

pydocstringformatter/formatting/base.py
DanielNoord: Let's do an attribute, it gives us a little more configurability while only requiring one additional lines per class.

pydocstringformatter/formatting/__init__.py
DanielNoord: Ah, understood! Never mind then.

pydocstringformatter/utils/argument_parsing.py
DanielNoord: Ah 😅 

Smart!

pydocstringformatter/utils/argument_parsing.py
DanielNoord: Hmm I can see this becoming very complicated.
What if I do:
`pydocstringformatter file --closing-quotes --no-closing-quotes` and `pydocstringformatter file --no-closing-quotes --closing-quotes`?

Pierre-Sassoulas: > What if I do: ``pydocstringformatter file --closing-quotes --no-closing-quotes`` and ``pydocstringformatter file --no-closing-quotes --closing-quotes`` ?

Well that would depends on argparse implementation detail. But I think this is an edge case that will never happen if a sentient being is using the program, there's no valid reason to use both. Another way to look at this is that if the parsing args depends on the value of optional, we have a breaking change in our arguments when we change the value of optional. Imo it's better to define both all the time so the API is stable (and the user can always expect both so the args are predictables).

DanielNoord: It would become a non-edge case if we made `--fix-quotes` which would do `--fix-closing-quotes` and `--fix-opening-quotes` and then have:
`--fix-quotes --no-fix-closing-quotes`.

This is still a bit silly, but I can impinge creating a preset of fixes (possibly `--strict`) of which you might then want to disable one or two. This is a similar issue as in `pylint` where the disabling of messages is dependent on the order of `enable` and `disable`. See https://github.com/PyCQA/pylint/issues/3696. I think that that is actually less intuitive for an end-user.

What about ``--closing-quotes=y`` or ``--closing-quotes=n``? Is possible I would even try to make ``--closing-quotes`` default to ``--closing-quotes=y``.
This does not fully solve the issue of ``--strict`` and ``--closing-quotes=n`` but I'd say we would want an implementation where specific check options suppress any "preset" options.

Pierre-Sassoulas: > What about --closing-quotes=y or --closing-quotes=n?

I don't like having to determine if we need an equal and what the truthey and falsey values are for this kind of options. I.e. What do you need to choose between ``--closing-quotes=y`` ``--closing-quotes y``, ``--closing-quotes=1`` or ``--closing-quotes=True``, etc.. So I use ``--x`` and ``--no-x`` which are shown in the --help and are explicit. But It's another solution, it would works too.

Pierre-Sassoulas: We can also make ``--closing-quotes`` ``--no-closing-quotes`` options mutually exclusive .

Pierre-Sassoulas: Actually the POSIX standard is pretty clear about that, I forgot to make it mutually exclusive. It's been a long time since I worked a fresh new lib 😄

> Guideline 11:
>     The order of different options relative to one another should not matter, unless the options are documented > as mutually-exclusive and such an option is documented to override any incompatible options preceding it. If an > option that has option-arguments is repeated, the option and option-argument combinations should be interpreted in the order specified on the command line.

DanielNoord: I agree, but having two separate options makes it so much harder to implement and determine what the users wants.
If ``closing-quotes`` default is `True` and due to presets ``closing-quotes`` is `True` and ``no-closing-quotes`` is also `True`, which do we follow?
If ``closing-quotes`` default is `False` and due to presets ``closing-quotes`` is `True` and ``no-closing-quotes`` is also `True`, which do we follow?

It would be much easier if we could have a single option that stored the state of the formatter. What we're looking for is `argparse.BooleanOptionalAction`, but that was only added in `3.9`.

DanielNoord: > Actually the POSIX standard is pretty clear about that, I forgot to make it mutually exclusive. It's been a long time since I worked on a fresh new lib 😄
> 
> > Guideline 11:
> > The order of different options relative to one another should not matter, unless the options are documented as mutually-exclusive and such an option is documented to override any incompatible options preceding it. If an option that has option-arguments is repeated, the option and option-argument combinations should be interpreted in the order specified on the command line.

We would then need to make ``no-closing-quotes`` exclusive with ``--strict`` which would be defeating the purpose of having presets on which users can then build their specific preferences.

DanielNoord: Looking at the implementation of `argparse.BooleanOptionalAction` we could actually try and see if we can copy it and use it ourselves on `3.8`? It doesn't look that difficult to maintain and would solve all of our troubles (I hope..)

Pierre-Sassoulas: > If closing-quotes default is True and due to presets closing-quotes is True and no-closing-quotes is also True, which do we follow?

Because of ` dest=name,` the two options all store in the same boolean (`closing-quotes`), so in the case you described, the default value is erased and `--no-closing-quote` store False in `closing-quotes`.

DanielNoord: Ah shit, I missed that. Hmm, then this might indeed be fine. It still feels a bit weird to see `store_true` and then a `default` argument.

Do we know what happens when we call `ArgumentParser.set_defaults`? Will the `store_true` option and its default option or the `store_false` be taken for `dest=name`?

DanielNoord: > It's been a long time since I worked on a fresh new lib 😄

Same btw, and because of all the issues we're having with configuration parsing in `pylint` (and other libraries) I am bit hesitant about changing anything. I want to be sure we're not missing anything 😄

pydocstringformatter/utils/argument_parsing.py
Pierre-Sassoulas: Done in latest commit

pydocstringformatter/utils/argument_parsing.py
Pierre-Sassoulas: >  I am bit hesitant about changing anything. 

In Pylint I would say the issue is optparse second and having callback first. It's a mess. But I think getting information from the Formatters (Checkers for pylint) is fine. Right now it's just name but we could imagine formatter specific option later. Getting global option from ``_register_arguments`` (somewhere in PyLinter for pylint) is also fine.

Pierre-Sassoulas: > .rc file for each directory? The latter would means less duplication.

It would also means more directories if we still want to duplicate because we need to :smile:  tbh I think both are fine

pydocstringformatter/formatting/__init__.py
```diff
@@ -1,11 +1,14 @@
-__all__ = ["FORMATTERS"]
+__all__ = ["FORMATTERS", "Formatter", "StringFormatter"]
```
Pierre-Sassoulas, line 1: I made the Base class public here @DanielNoord

DanielNoord: Perhaps we should wait with that for `1.0.0`? Make sure our internal APIs are all stable and then decide what we want and what not to expose?

tests/test_formatter.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+
+from pydocstringformatter.formatting import FORMATTERS
+from pydocstringformatter.formatting.base import Formatter
+
+
+@pytest.mark.parametrize(
+    "formatter,expected", zip(FORMATTERS, ["beginning-quotes", "closing-quotes"])
+)
+def test_name(formatter: Formatter, expected: str) -> None:
```
DanielNoord, line 10: This might be hard to maintain at some point.

What about:
```python
formatter_names: Set[str] = {}
for formatter in FORMATTERS:
    assert formatter.name, "Each formatter should have a name set."
    assert formatter.name not in formatter_names, "Each formatter should have an unique name."
    formatter_names.add(formatter.name)
```

We could even separate the `name` property and this test into a separate PR to keep a clean commit history.

Pierre-Sassoulas: Right, now that the name isn't generated there's no need to make sure that someone check that the "name" of the new formatter make sense.

DanielNoord: 👍 

No, but an exclusivity check makes sense I think. Wouldn't hurt in any case.

pydocstringformatter/formatting/__init__.py
Pierre-Sassoulas: Done

tests/test_formatter.py
Pierre-Sassoulas: Done

pydocstringformatter/utils/argument_parsing.py
Pierre-Sassoulas: Should we do the test refactor in another MR ? Maybe in this MR ?

DanielNoord: Yeah I think so.

However, we might want to wait on #2 as that would drastically change the output of each tests. With `difflib` which is in `stdlib` it should be fairly trivial to generate `diffs`, but I haven't got round to it (and I'm quite busy this week).
I think because of file names it will be impossible to create `.py.out` with correct diffs, so we might need to move the "functional" tests to checking the written state of a temporary file passed to the test runner?
Lots to do still 😄

Pierre-Sassoulas: OK, I'll let you do the diff change, I'm not familiar with this lib. Sorry for being a little too enthusiastic about this fresh project and making you review a lot of code on a busy week 😄

tests/test_formatter.py
Pierre-Sassoulas: Opened up #18 and changed the branch we want to merge in here :)

.github/CONTRIBUTING.md
```diff
@@ -8,7 +8,8 @@ Use `pre-commit install` to install the pre-commit hook for the repository.
 
 - Implement a Formatter by inheriting from ``pydocstringformatter.formatting.Formatter``
 - Add your new formatter to ``pydocstringformatter.formatting.FORMATTERS``
-- Choose a proper name because this will be user-facing: the name will be used for options of the CLI.
+- Create a clear docstring because this will be user-facing: it's what will be seen in the help.
+- Choose a proper name because this will be user-facing: the name will be used for option of the CLI.
```
DanielNoord, line 12:
```suggestion
- Write a clear docstring because this will be user-facing: it's what will be seen in the help message for the formatters command line option.
- Choose a proper name because this will be user-facing: the name will be used to turn the checker on and off via the command line or config files.
```

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
 def _register_arguments(version: str) -> argparse.ArgumentParser:
     """Create an argument parser and add all supported arguments"""
     parser = argparse.ArgumentParser(prog="pydocstringformatter")
```
DanielNoord, line 22: Could your re-add the white lines?

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
...
+) -> argparse.ArgumentParser:
+    """Register a list of formatters, so they can all be deactivated or activated."""
+    for formatter in formatters:
+        name = formatter.name
+        help_text = f"ctivate the {name} formatter"
+        parser.add_argument(
+            f"--{name}",
+            action="store_true",
+            dest=name,
+            default=not formatter.optional,
```
DanielNoord, line 51: I have thought this a little more and I think it would be better if we removed the `default` here and did another `_register_formatter_defaults` function. The documentation is pretty clear about what `store_true` is "supposed" to be used for:
`'store_true' and 'store_false' - These are special cases of 'store_const' used for storing the values True and False respectively. In addition, they create default values of False and True respectively`.

Just to be sure for future changes I think it makes more sense to honour the default values created by `store_true` and then parse the `optional` attribute. That also makes more sense for any outsider looking at this code for the first time.

What do you think?

pydocstringformatter/run.py
```diff
@@ -9,13 +9,17 @@
...
 from pydocstringformatter import __version__, formatting, utils
+from pydocstringformatter.formatting import FORMATTERS
 
 
 class _Run:
     """Main class that represent a run of the program"""
 
     def __init__(self, argv: Union[List[str], None]) -> None:
         self.arg_parser = utils._register_arguments(__version__)
+        self.arg_parser = utils._register_arguments_formatters(
```
DanielNoord, line 20: I believe this doesn't actually need to be re-assigned, as the object is changed in place by `add_argument`. Just a very small optimisation 😄

DanielNoord (changes requested):

pydocstringformatter/utils/argument_parsing.py
Pierre-Sassoulas: I think I don't understand exactly what you propose. Could you adapt what we currently have ?
```python
parser.add_argument(
    f"--{name}",
    action="store_true"
    default=not optional,
)
parser.add_argument(
    f"--{name}",
    action="store_false",
)    
```          
It seems that if we remove the default value then we need additional parsing later on ? Can we guess if the value is user defined or simply the default ?

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
 def _register_arguments(version: str) -> argparse.ArgumentParser:
     """Create an argument parser and add all supported arguments"""
     parser = argparse.ArgumentParser(prog="pydocstringformatter")
-
     parser.add_argument("files", nargs="*", type=str)
```
Pierre-Sassoulas, line 23:
```suggestion

    parser.add_argument("files", nargs="*", type=str)

```

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -19,24 +20,40 @@ def _parse_command_line_arguments(
...
-
     parser.add_argument(
         "-w",
         "--write",
         action="store_true",
         help="Write the changes to file instead of printing the files to stdout",
     )
```
Pierre-Sassoulas, line 29:
```suggestion
    )

```

pydocstringformatter/utils/argument_parsing.py
DanielNoord: We'll need some sort of parsing later on anyway to make sure the `default` value actually gets into the `Namespace` object. I think it would be better to separate that parsing (and thus the parsing of the `Formatter.optional` attribute into its own function) than setting the default value here to a value that can be counter-intuitive to what the documentation says.

As for parsing and whether it is user defined I think this dependent on the order in which we parse.
I would suggest: defaults > config files > command line.

Pierre-Sassoulas: But right now we don't have to parse anything as the ``default`` override what ``store_const`` is doing (logically). If we create config file and need another layer we could use ``confuse`` to do something like this:
```python
config: confuse.Configuration
config_default = config[name].get(confuse.Optional("bool"))
default = not optional if config_default is None else config_default 
parser.add_argument(
    f"--{name}",
    action="store_true"
    default=default 
)
parser.add_argument(
    f"--{name}",
    action="store_false",
) 
```
Also this is a part of the code that can almost never change as it's decoupled and generic, we don't have to touch it for every new formatter.

pydocstringformatter/run.py
```diff
@@ -46,10 +48,11 @@ def _format_file(self, filename: Path) -> bool:
 
         for index, tokeninfo in enumerate(tokens):
             new_tokeninfo = tokeninfo
-
+            formatter_options = vars(self.config)
```
Pierre-Sassoulas, line 51:
```suggestion

            formatter_options = vars(self.config)
```

pydocstringformatter/utils/argument_parsing.py
DanielNoord: I have never worked with `confuse` so I don't know specifically what it does here.

I understand how the current format makes the code more efficient and achieve the same with less lines, but I think in the long run it only adds maintainability costs to the project: every peculiar thing a project does that is not "as expected" increases the cost for new contributors to understand what is happening. To me it makes much more sense to add two options using `store_false` and `store_true` and then parse a default value to store the correct value in `dest`. To me that feels more like the natural flow of `argparse`. However, if you're insistent on this we can keep your suggestion 😄

Pierre-Sassoulas: confuse is pretty nice let me show you with a confuse example :)
- Initializing the config : https://github.com/Pierre-Sassoulas/centralized-pre-commit-conf/blob/master/centralized_pre_commit_conf/main.py#L13
- confuse will search for the system configuration file in `/home/user/.config/pre-commit-conf` for Linux and
`%HOME%\AppData\Roaming\pre-commit-conf for Windows`... Then if it does not find any it uses the default
- Default value are here : https://github.com/Pierre-Sassoulas/centralized-pre-commit-conf/blob/master/centralized_pre_commit_conf/config_default.yaml
- Or if something is overridden in the configuration it uses the overridden value. The adapted argparse to do that using the confuse configuration : https://github.com/Pierre-Sassoulas/centralized-pre-commit-conf/blob/master/centralized_pre_commit_conf/parse_args.py#L21

So confuse handles the overriding of configuration (pretty intuitively, as it's what the lib does). There is a twitch for us : When do we recover our own default value from the formatter ? So we do not define a default value in our own yaml config ,so if confuse has something it must comes from the user. After that the following line: `default = not optional if config_default is None else config_default ` actually does "if the configuration from confuse is not specified than the default comes from our formatter".

DanielNoord: Hmm `confuse` looks nice, but is it really necessary for us? We don't really need to look for config files in `usr` directories (right now) and I think adding too many dependencies to such a simple auto-formatter might make it less compatible with other apps.

In the end I really like the approach of `--formatter` and `--no-formatter` (I didn't think of that myself), but I'd just like 1) not to overwrite the default value for a `store_true` option and 2) keep option adding and option parsing separate so we can easily iterate on them and know what is responsible for what.

Pierre-Sassoulas:
I added the last commit to be able to cover the case where an optional formatter was used (we don't have any in FORMATTERS yet). Previously it was handled in argparse with the default value so did not need to be tested. I guess it would permit to expose the Run function to the outside so user can define their own formatters if we want. I can't separate it from this MR easily but we could do it later if we find another way to cover the creation of argument for optional formatter.

tests/test_run.py
```diff
@@ -106,3 +115,55 @@ def test_output_message_two_files(
...
+
+
+@pytest.mark.parametrize(
+    "args,should_format",
+    [
+        [[f"--no-{f.name}" for f in FORMATTERS], False],
+        [[f"--{f.name}" for f in FORMATTERS], True],
+    ],
+)
+def test_optional_checker(
```
Pierre-Sassoulas, line 149:
```suggestion
def test_optional_formatters(
```

.github/CONTRIBUTING.md
```diff
@@ -8,8 +8,10 @@ Use `pre-commit install` to install the pre-commit hook for the repository.
 
 - Implement a Formatter by inheriting from `pydocstringformatter.formatting.Formatter`
 - Add your new formatter to `pydocstringformatter.formatting.FORMATTERS`
-- Choose a proper name because this will be user-facing: the name will be used for
-  options of the CLI.
+- Write a clear docstring because this will be user-facing: it's what will be seen in
+  the help message for the formatters command line option.
```
DanielNoord, line 12:
```suggestion
  the help message for the formatter's command line option.
```

.github/CONTRIBUTING.md
```diff
@@ -8,8 +8,10 @@ Use `pre-commit install` to install the pre-commit hook for the repository.
...
+- Choose a proper name because this will be user-facing: the name will be used to turn
+  the checker on and off via the command line or config files.
```
DanielNoord, line 14:
```suggestion
  the formatter on and off via the command line or config files.
```

Little `pylint` slipping in here? 😄

pydocstringformatter/run.py
```diff
@@ -6,22 +6,27 @@
 import sys
 import tokenize
 from pathlib import Path
-from typing import List, Union
+from typing import List, Optional, Union
 
-from pydocstringformatter import __version__, formatting, utils
+from pydocstringformatter import __version__, utils
+from pydocstringformatter.formatting import FORMATTERS, Formatter
```
DanielNoord, line 12: Personally I like the `formatting` import as it helps show the namespace of stuff, but I'm not too opinionated about this. If you prefer this I'm fine with it!

pydocstringformatter/run.py
```diff
@@ -6,22 +6,27 @@
...
 
 
 class _Run:
     """Main class that represent a run of the program."""
 
-    def __init__(self, argv: Union[List[str], None]) -> None:
+    def __init__(
+        self, argv: Union[List[str], None], formatters: Optional[List[Formatter]] = None
```
DanielNoord, line 19: Allowing additional `formatters` should probably go in its own PR

pydocstringformatter/run.py
```diff
@@ -47,9 +52,11 @@ def _format_file(self, filename: Path) -> bool:
         for index, tokeninfo in enumerate(tokens):
             new_tokeninfo = tokeninfo
 
+            formatter_options = vars(self.config)
             if utils._is_docstring(new_tokeninfo, tokens[index - 1]):
-                for formatter in formatting.FORMATTERS:
-                    new_tokeninfo = formatter.treat_token(new_tokeninfo)
+                for formatter in self.formatters:
+                    if formatter_options[formatter.name]:
```
DanielNoord, line 58: Can't we do `self.config.formatter`? That would save the assignment. I believe `Namespace` is special-cased by `mypy` not to generate any issues about this.

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -80,3 +99,37 @@ def _parse_toml_file(
...
 
         parser.parse_args(arguments, namespace)
+
+
+def _load_default_formatters(
+    parser: argparse.ArgumentParser,
+    namespace: argparse.Namespace,
+    formatters: List[Formatter],
+) -> None:
+    """Load the default formatters based on their 'optional' attribute."""
```
DanielNoord, line 109:
```suggestion
    """Load the list of formatters based on their 'optional' attribute."""
```

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -80,3 +99,37 @@ def _parse_toml_file(
             arguments += _parse_toml_option(key, value)
 
         parser.parse_args(arguments, namespace)
+
+
+def _load_default_formatters(
```
DanielNoord, line 104:
```suggestion
def _load_formatters_default_option(
```

If we can indeed pass non-standard formatter this is not longer only `default`.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ py-version="3.8"
 disable = [
     "missing-module-docstring",
     "too-few-public-methods",
+    "no-self-use",
```
DanielNoord, line 36: Shouldn't we decorate with `staticmethod` instead?

DanielNoord (changes requested):
Almost there!

Personally I would be okay with passing this with the decreased coverage and not adding the `non-standard formatter` option just yet. I'd like to make #25 optional to begin with, as I think it might be quite "opinionated" and would require some refactoring for most projects. (For example, I wonder how strictly we currently adhere to this in `pylint`).
When I implement #25 I could add tests for `optional = True` cases and increase the coverage to 100% again. 

Let me know what you think!

DanielNoord:
Btw, after this we should probably create a follow-up issue to create something like `enable`, `disable` in `pyproject.toml`.

.github/CONTRIBUTING.md
Pierre-Sassoulas: Haha, yeah muscle memory... I fixed the function name just before 😄

pyproject.toml
Pierre-Sassoulas: I think this is an annoying pylint's message that we're going to make optional soon, but yeah that's the fix 😄

DanielNoord: But this isn't a false positive right?

pydocstringformatter/run.py
Pierre-Sassoulas: I think if we do that we must control the formatters through the option parsing only and it would be harder to use the ``Run`` interface.

pydocstringformatter/run.py
Pierre-Sassoulas: I started creating another MR, but it's embedded with the way we changed the formatter to become optional, thus hard to cherry-pick.

pydocstringformatter/run.py
Pierre-Sassoulas: Done

Pierre-Sassoulas:
> I would be okay with passing this with the decreased coverage

That would be a way to not have to mix the change in Run's API with this one. I think it's a configuration in coverage to change ?

pyproject.toml
Pierre-Sassoulas: I think it happened when ``def treat_token(self, tokeninfo: TokenInfo) -> TokenInfo:`` in the OptionalFormatter wasn't the same than in the base class (during dev), it disappeared now.

DanielNoord:
> > I would be okay with passing this with the decreased coverage
> 
> That would be a way to not have to mix the change in Run's API with this one. I think it's a configuration in coverage to change ?

Yeah, I set it to disallow any negative change in coverage. But since it isn't a required action I think you can merge while the `coverage` actions fails right? I would be fine with just letting coverage drop on this one. Alternatively you could add a `TODO` linking to #25 and set a `pragma: no cover`.

pydocstringformatter/run.py
DanielNoord: Really? We would still call `utils._register_arguments_formatters` so any non-standard formatters get registered to the `ArgumentParser` (if we every extend the `Run` API to allow that, like I said probably not in this PR).
I just referred to this line. Instead of using `vars` to store the `Namespace` as a dict can't we just use `formatter.name` as the key on the `Namespace` object (stored on `self.config` directly)?

pydocstringformatter/run.py
DanielNoord: Thanks 😄

pydocstringformatter/run.py
Pierre-Sassoulas: Ho sorry, I thought that we were talking about ``self.formatter`` vs ``self.config.formatter``. Too much non ascii checker review this week !

.github/CONTRIBUTING.md
```diff
@@ -8,8 +8,10 @@ Use `pre-commit install` to install the pre-commit hook for the repository.
 
 - Implement a Formatter by inheriting from `pydocstringformatter.formatting.Formatter`
 - Add your new formatter to `pydocstringformatter.formatting.FORMATTERS`
-- Choose a proper name because this will be user-facing: the name will be used for
-  options of the CLI.
+- Write a clear docstring because this will be user-facing: it's what will be seen in
```
DanielNoord, line 11: I just thought about this. What do we do with the help message if the docstring contains both a summary and body?
Not something to be fixed in this PR perhaps, but we might want to add a `TODO` and open an issue to track.

pydocstringformatter/run.py
Pierre-Sassoulas:
```
>                   if self.config["formatter"][formatter.name]:
E                   TypeError: 'Namespace' object is not subscriptable
```
```
>                   if self.config.formatter[formatter.name]:
E                   AttributeError: 'Namespace' object has no attribute 'formatter'
```

Calling vars multiple time seem like a performance issue ?

Pierre-Sassoulas: Nvm we can do ``if getattr(self.config, formatter.name):``

DanielNoord: Hmm, what about:
```python
self.formatters = {i.name: i for i in formatting.FORMATTERS}

# and then

for formatter_name in self.formatter:
    if self.config.formatter_name:
        # do something
```

I'm not totally sure what type of keys `Namespace` allows, but it looked weird to call `vars` on an object that is basically a `dict` itself already.

.github/CONTRIBUTING.md
Pierre-Sassoulas: See #28

pydocstringformatter/run.py
Pierre-Sassoulas: We need to use ``getattr(self.config, formatter.name):`` if we don't want to use ``vars``

DanielNoord: Fine with me!

tests/test_run.py
```diff
@@ -106,3 +113,33 @@ def test_output_message_two_files(
...
+    args: List[str],
+    should_format: bool,
+    capsys: pytest.CaptureFixture[str],
+    tmp_path: Path,
+) -> None:
+    """Test that optional check are activated or not depending on options."""
+    bad_docstring = tmp_path / "bad_docstring.py"
+    bad_docstring.write_text(f'"""{"a" * 120}\n{"b" * 120}"""')
+    with patch.object(sys, "argv", ["pydocstringformatter", str(bad_docstring)] + args):
+        pydocstringformatter.run_docstring_formatter()
```
DanielNoord, line 135:
```suggestion
    pydocstringformatter.run_docstring_formatter([str(bad_docstring) + args])
```

tests/test_run.py
```diff
@@ -106,3 +113,33 @@ def test_output_message_two_files(
...
+    out, err = capsys.readouterr()
+    assert not err
+    if should_format:
+        msg = "Nothing was modified, but all formatters are activated."
+        assert "Nothing to do!" not in out
+        expected = ["---", "@@", "+++"]
+        assert all(e in out for e in expected), msg
```
DanielNoord, line 142: This is quite a neat way to test changes occurred! Might use this myself for future tests as well!

tests/test_run.py
```diff
@@ -106,3 +113,33 @@ def test_output_message_two_files(
...
+        [[f"--{f.name}" for f in FORMATTERS], True],
+    ],
+)
+def test_optional_formatters(
+    args: List[str],
+    should_format: bool,
+    capsys: pytest.CaptureFixture[str],
+    tmp_path: Path,
+) -> None:
+    """Test that optional check are activated or not depending on options."""
```
DanielNoord, line 131:
```suggestion
    """Test that (optional) formatters are activated or not depending on options."""
```

tests/test_run.py
```diff
@@ -2,19 +2,26 @@
...
+from unittest.mock import patch
 
 import pytest
 
 import pydocstringformatter
+from pydocstringformatter.formatting import FORMATTERS
 
 
 def test_no_arguments(capsys: pytest.CaptureFixture[str]) -> None:
     """Test that we warn when no arguments are provided"""
```
DanielNoord, line 15:
```suggestion
    """Test that we display a help message when no arguments are provided."""
```

tests/test_run.py
```diff
@@ -2,19 +2,26 @@
...
     sys.argv = ["pydocstringformatter"]
     pydocstringformatter.run_docstring_formatter()
-    output = capsys.readouterr()
-    assert output.out.startswith("usage: pydocstringformatter [-h]")
-    assert not output.err
+    out, err = capsys.readouterr()
+    assert out.startswith("usage: pydocstringformatter [-h]")
+    assert "--beginning-quotes" in out
```
DanielNoord, line 20:
```suggestion
    
    # Test that we print help messages for individual formatters as well
    assert "--beginning-quotes" in out
```

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -80,3 +99,37 @@ def _parse_toml_file(
...
 
         parser.parse_args(arguments, namespace)
+
+
+def _load_formatters_default_option(
+    parser: argparse.ArgumentParser,
+    namespace: argparse.Namespace,
+    formatters: List[Formatter],
+) -> None:
+    """Load the list of formatters based on their 'optional' attribute."""
```
DanielNoord, line 109:
```suggestion
    """Parse the state of a the list of formatters based on their 'optional' attribute."""
```

DanielNoord (approved):
Final changes!

Thanks @Pierre-Sassoulas. I was quite insistent on some changes in this PR, but I'm really happy with the final system you/we came up with. This feels like it really suits our needs but is also easily expandable. Thanks for your help!

DanielNoord:
Oops, I made a typo in my suggestion. Sorry about that!

pydocstringformatter/utils/argument_parsing.py
```diff
@@ -80,3 +99,37 @@ def _parse_toml_file(
...
 
         parser.parse_args(arguments, namespace)
+
+
+def _load_formatters_default_option(
+    parser: argparse.ArgumentParser,
+    namespace: argparse.Namespace,
+    formatters: List[Formatter],
+) -> None:
+    """Parse the state of a the list of formatters based on their 'optional' attribute."""
```
DanielNoord:
```suggestion
    """Parse the state of a list of formatters based on their 'optional' attribute."""
```

Pierre-Sassoulas:
Good design need some heated discussion me think 😄

pydocstringformatter/utils/argument_parsing.py
Pierre-Sassoulas: Had to apply that locally anyway because the docstring was too long for flake8 :)

PR #16, opened 2022-01-04
Reconfigure encoding of ``stdout`` to ensure emoji's can be displayed

### LINKED ISSUE #13: Emoji in run.py crashes using pre-commit python 3.8.10
Hey Daniel,
The emoji in the last `print` statement under `run.py` crashes when using Python version that does not support emojis.

`UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f389' in position 42: character maps to <undefined>`

For example for my `Python 3.5.3`, Windows 10 Pro 64 bit.

Do we want to support older version of Python with this tool? If so maybe the print statement should be modified?

DanielNoord:
Closes #13

### DIFF
--- a/pydocstringformatter/utils/output.py
+++ b/pydocstringformatter/utils/output.py
@@ -1,0 +1,20 @@
+import sys
+
+
+def _encode_string(string: str) -> bytes:
+    """Encode a string to utf-8.
+
+    This can be used to circumvent the issue of the standard encoding
+    of a windows console not being utf-8.
+    See: https://github.com/DanielNoord/pydocstringformatter/issues/13
+    """
+    return string.encode("utf-8")
+
+
+def _print_to_console(string: str) -> None:
+    """Print a string to the console while handling edge cases.
+
+    This should be used instead of print() whenever we want to
+    print emoji's or non-ASCII characters.
+    """
+    sys.stdout.buffer.write(_encode_string(string))
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,7 +1,11 @@
+import sys
 import tokenize
 from pathlib import Path
 from typing import List, Tuple
 
+import pytest
+
+import pydocstringformatter
 from pydocstringformatter.utils import _find_python_files, _is_docstring
 
 HERE = Path(__file__)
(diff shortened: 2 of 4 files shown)


Pierre-Sassoulas (commented):
It's hard to test, but maybe we could set (the|some] windows jobs with the problematic encoding ?

DanielNoord:
> It's hard to test, but maybe we could set (the|some] windows jobs with the problematic encoding ?

That would be akin to:
```python
sys.stdout.reconfigure(encoding="cp1521")
sys.stdout.reconfigure(encoding="utf-8")
assert sys.stdout.encoding == "utf-8"
```

Feels a bit silly 😅

Pierre-Sassoulas:
I meant the CI job directly with a system command, before launching the tests.

DanielNoord:
Ah, that might work! Let's wait to see if we get any other good ideas for adding this to the unittests directly.

DanielNoord:
https://github.com/pre-commit/pre-commit-hooks/pull/571/files

This might also work. We can then just check encoding of all messages written to `stdout` and see if they are indeed `uff-8`?

Pierre-Sassoulas:
It look like one of the highest answer of the stackoverflow question linked in the original issue (https://stackoverflow.com/a/27093194/2519059)

orSolocate:
Hey guys I reproduced the error locally and verified that your suggested addition solves it!

DanielNoord:
@orSolocate I think I actually wrote a test case that test the real-world scenario but would you able to report back if this new fix also works?

pydocstringformatter/utils/output.py
```diff
@@ -0,0 +1,20 @@
...
+    """Encode a string to utf-8.
+
+    This can be used to circumvent the issue of the standard encoding
+    of a windows console not being utf-8.
+    See: https://github.com/DanielNoord/pydocstringformatter/issues/13
+    """
+    return string.encode("utf-8")
+
+
+def _print_to_console(string: str) -> None:
```
Pierre-Sassoulas, line 14:
```suggestion
def _print_with_emoji(string: str) -> None:
```
The probability of it being useless for something else is slim 😄

tests/test_utils.py
```diff
@@ -133,3 +137,22 @@ def test_module_docstrings(self) -> None:
...
+
+def test_encoding_of_console_messages(
+    capsys: pytest.CaptureFixture[str], test_file: str
+) -> None:
+    """Test that we can print emoji's to non utf-8 consoles.
+
+    Regression test for:
+    https://github.com/DanielNoord/pydocstringformatter/issues/13
+    """
+    sys.stdout.reconfigure(encoding="cp1252")  # type: ignore[attr-defined]
```
Pierre-Sassoulas, line 150: Yeap, that's pretty elegant, good job !

Pierre-Sassoulas (approved):

pydocstringformatter/utils/output.py
DanielNoord: You'll be surprised: I think the implementation of this in `click` is 100+ lines long 😅 

They handle every possible edge case, but I thought it would be good to separate every little thing into its own function if we ever need to expand on it.

orSolocate:
@DanielNoord  Works!
verified it crashes without it and passed with it. You should push it

DanielNoord:
Thanks @orSolocate!

PR #17, opened 2022-01-04
Create a file for the base classes for Formatters

Pierre-Sassoulas:
Refactor prior to #15

### DIFF
--- a/pydocstringformatter/formatting/base.py
+++ b/pydocstringformatter/formatting/base.py
@@ -3,19 +3,19 @@ import tokenize
 
 
 class Formatter:
-    """Base class for docstring formatter"""
+    """Base class for docstring formatter."""
 
     @abc.abstractmethod
     def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
-        """Return a modified token"""
+        """Return a modified token."""
 
 
 class StringFormatter(Formatter):
-    """Base class for formatter that only modifies the string content"""
+    """Base class for formatter that only modifies the string content."""
 
     @abc.abstractmethod
     def _treat_string(self, tokeninfo: tokenize.TokenInfo) -> str:
-        """Return a modified string"""
+        """Return a modified string."""
 
     def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
         return tokenize.TokenInfo(
--- a/pydocstringformatter/formatting/base.py
+++ b/pydocstringformatter/formatting/base.py
@@ -1,0 +1,27 @@
+import abc
+import tokenize
+
+
+class Formatter:
+    """Base class for docstring formatter"""
+
+    @abc.abstractmethod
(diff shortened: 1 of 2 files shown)


DanielNoord (commented):
After all the work you did I need to take another good look at what is private and isn't, but I think it would make sense to expose these two classes in the public API.

pydocstringformatter/formatting/base.py
```diff
@@ -0,0 +1,27 @@
+import abc
+import tokenize
+
+
+class Formatter:
+    """Base class for docstring formatter"""
```
DanielNoord, line 6:
```suggestion
    """Base class for docstring formatter."""
```

pydocstringformatter/formatting/base.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @abc.abstractmethod
+    def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+        """Return a modified token"""
```
DanielNoord, line 10:
```suggestion
        """Return a modified token."""
```

pydocstringformatter/formatting/base.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class StringFormatter(Formatter):
+    """Base class for formatter that only modifies the string content"""
+
+    @abc.abstractmethod
+    def _treat_string(self, tokeninfo: tokenize.TokenInfo) -> str:
+        """Return a modified string"""
```
DanielNoord, line 18:
```suggestion
        """Return a modified string."""
```

pydocstringformatter/formatting/base.py
```diff
@@ -0,0 +1,27 @@
...
+class Formatter:
+    """Base class for docstring formatter"""
+
+    @abc.abstractmethod
+    def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+        """Return a modified token"""
+
+
+class StringFormatter(Formatter):
+    """Base class for formatter that only modifies the string content"""
```
DanielNoord, line 14:
```suggestion
    """Base class for formatter that only modifies the string content."""
```

DanielNoord (approved):

pydocstringformatter/formatting/base.py
Pierre-Sassoulas: Huhu, you can't wait for this new DotFormatter huh 😄

DanielNoord: Well, I'm actually a bit hesitant. I would like to do something like "CheckIfSentenceFormatter" first, as adding a dot to non-sentences is weird and might shy users away from using the tool.

Perhaps I can leverage the work you're doing in #15 to make it an optional formatter to begin with 😄

Pierre-Sassoulas: Yes it's non trivial, I think a first version could add a line feed after existing dot and add a dot to string not ending with ``:``.

PR #19, opened 2022-01-04
Print out ``diff`` instead of complete file for formatting changes

### LINKED ISSUE #2: Print out ``diffs`` instead of complete files

DanielNoord:
Closes #2 

It's a bit annoying to have to check for `diffs` in the non-formatting unittests, but I think we can manage. If the `endswith` becomes too long we can always do `starswith("---")` although the `endswith` also provides a good sanity check that nothing is wrong with the actual diff.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -17,7 +17,9 @@ def test_no_toml(
     monkeypatch.chdir(CONFIG_DATA / "no_toml")
     pydocstringformatter.run_docstring_formatter(["test_package"])
     output = capsys.readouterr()
-    assert output.out == '"""A docstring"""\n'
+    assert output.out.endswith(
+        '@@ -1,3 +1,2 @@\n-"""\n-A docstring"""\n+"""A docstring"""\n '
+    )
     assert not output.err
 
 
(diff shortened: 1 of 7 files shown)


tests/test_config.py
```diff
@@ -17,7 +17,9 @@ def test_no_toml(
     monkeypatch.chdir(CONFIG_DATA / "no_toml")
     pydocstringformatter.run_docstring_formatter(["test_package"])
     output = capsys.readouterr()
-    assert output.out == '"""A docstring"""\n'
+    assert output.out.endswith(
+        '@@ -1,3 +1,2 @@\n-"""\n-A docstring"""\n+"""A docstring"""\n '
```
Pierre-Sassoulas, line 21: Should probably be using ```'''```, sorry I can't suggest, I'm on mobile

tests/test_formatting.py
```diff
@@ -25,10 +26,29 @@
     TESTS,
     ids=TEST_NAMES,
 )
-def test_formatting(test_file: str, capsys: pytest.CaptureFixture[str]) -> None:
-    """Test that we correctly format all files in the format directory"""
-    pydocstringformatter.run_docstring_formatter([test_file])
+def test_formatting(
+    test_file: str, capsys: pytest.CaptureFixture[str], tmp_path: pathlib.Path
+) -> None:
+    """Test that we correctly format all files in the format directory.
```
Pierre-Sassoulas, line 32: Very nice change !

Pierre-Sassoulas (approved):
Looks good, I would only change the tests string so it's easier to write ans understand.

Pierre-Sassoulas (approved):
👌

PR #21, opened 2022-01-06
Make every formatter optional 2

DanielNoord:
WIP.

Just to show how I would do the changes proposed in #15.

@Pierre-Sassoulas I'd still like to merge L19 and L20 (and possible create an `ArgumentManager` class to handle all of this a bit better), but with this approach (to me) it's much clearer where all options come from and where they get their value. By doing the default value parsing just before parsing all other configuration sources you know where any expected values could come from.


Pierre-Sassoulas (commented):
I see what you want to do now. It would work but I think using confuse (and as long as we have no config argparse's default argument) would be less verbose and require less code to maintain on our side. Did you check the confuse :) ?

Pierre-Sassoulas:
Closing after cherry-picking and rebasing on #15

PR #23, opened 2022-01-06
Add formatter of final period for single line docstrings and summaries

### LINKED ISSUE #10: Enforce a dot at the end of single line docstrings and summaries
This is discussed in [PEP 257](https://www.python.org/dev/peps/pep-0257/)

DanielNoord:
Pretty large change due to changes to the existing tests and codebase. Should probably be reviewed via the commit view.

README change explains pretty well what this does. Neat little feature for `0.3.0` I think :)

This closes #10

### DIFF
--- a/pydocstringformatter/formatting/__init__.py
+++ b/pydocstringformatter/formatting/__init__.py
@@ -6,9 +6,11 @@ from pydocstringformatter.formatting.base import Formatter
 from pydocstringformatter.formatting.formatter import (
     BeginningQuotesFormatter,
     ClosingQuotesFormatter,
+    FinalPeriodFormatter,
 )
 
 FORMATTERS: List[Formatter] = [
     BeginningQuotesFormatter(),
     ClosingQuotesFormatter(),
+    FinalPeriodFormatter(),
 ]
--- a/tests/data/format/final_period/function_docstrings.py
+++ b/tests/data/format/final_period/function_docstrings.py
@@ -1,0 +1,29 @@
+def func()
+    """Docstring"""
+
+    def inner_func()
+        """Docstring"""
+
+
+def func()
+    """A multi-line
+    docstring
+    """
+
+    def inner_func()
+        """A multi-line
+        docstring
+        """
+
+
+def func()
+    """Summary
+
+    docstring
+    """
+
+    def inner_func()
+        """Summary
+
+        docstring
+        """
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ multi-line docstring
 """
 ```
 
-**PEP 256: _The closing quotes are on the same line as the opening quotes_**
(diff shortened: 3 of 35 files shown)


pydocstringformatter/formatting/__init__.py
```diff
@@ -6,9 +6,11 @@
 from pydocstringformatter.formatting.formatter import (
     BeginningQuotesFormatter,
     ClosingQuotesFormatter,
+    FinalPeriodFormatter,
 )
 
 FORMATTERS: List[Formatter] = [
     BeginningQuotesFormatter(),
     ClosingQuotesFormatter(),
+    FinalPeriodFormatter(),
```
Pierre-Sassoulas, line 15:
```suggestion
        FinalPeriodFormatter(),
        ClosingQuotesFormatter(),
```
So we manage the line break if we go further than the alloted limit with the dot ?

tests/data/format/final_period/function_docstrings.py
```diff
@@ -0,0 +1,29 @@
...
+def func()
+    """A multi-line
+    docstring
+    """
+
+    def inner_func()
+        """A multi-line
+        docstring
+        """
+
```
Pierre-Sassoulas, line 17:
```suggestion

    def other_inner():
        """A docstring with a sentence ending with dot. And another one
        continuing later."""
```

I think the result should be:
```python
    def other_inner():
        """A docstring with a sentence ending with dot.

        And another one continuing later."""
```

Pierre-Sassoulas (changes requested):

pydocstringformatter/formatting/__init__.py
DanielNoord: Hmm, but I depend on the number of line breaks to figure out what to do..

I think for now this is fine, until we have a good way of doing line wrapping? After which we'll probably need to update the order?

tests/data/format/final_period/function_docstrings.py
DanielNoord: I didn't want to do too much with line wrapping and breaking in this and only focus on adding periods when we 99% know that they should be there (I skipped the bodies of multi line strings for the same reason). But I agree this should be done at some point.

What about `SplitSummaryAndDocstringFormatter` which splits on periods? That should be relatively easy to implement.

Pierre-Sassoulas: https://github.com/DanielNoord/pydocstringformatter/issues/25

pydocstringformatter/formatting/__init__.py
Pierre-Sassoulas: Fair enough !

Pierre-Sassoulas (approved):

PR #29, opened 2022-01-08
Add an external formatter API to be able to launch Run with custom formatters

Pierre-Sassoulas:
Was necessary to tests the args parsing with the default value handled by something else than argparse. As we decided in #15 to let the coverage fall a little we could still use it to provide an API later on.


DanielNoord:
I'm converting this to draft as I wonder if during development towards `1.0` will find other necessary attributes and methods needed for formatters. For example, a way to set formatter specific options.

I think this would be a really good feature to put in `1.1`. As in: make sure we get `1.0` to be compliant with both PEPs and that we are happy with how everything is set up, and then allow for some additional customisation with plugin formatters.

PR #31, opened 2022-01-08
Add an ``exclude`` option to exclude files based on ``glob`` patterns

### LINKED ISSUE #1: Add an ``exclude`` option
Should be able to exclude certain files or paths.

DanielNoord:
Closes #1 

`glob` is apparently pretty bad for ignoring stuff and is only used to capture things. (include vs. exclude).

However, for now I think this will be fine. I'm not sure about the tests and whether this catches everything, but at least we can get feedback if this works now!

### DIFF
--- a/tests/data/config/exclude_match/pyproject.toml
+++ b/tests/data/config/exclude_match/pyproject.toml
@@ -1,0 +1,2 @@
+[tool.pydocstringformatter]
+exclude = "**/test_package/**"
--- a/README.md
+++ b/README.md
@@ -19,21 +19,32 @@ pip install pydocstringformatter
 ## Usage
 
 ```shell
-usage: pydocstringformatter [-h] [-w] [-v] [files ...]
+usage: pydocstringformatter [-h] [-w] [--exclude EXCLUDE] [-v] [files ...]
 
 positional arguments:
   files
 
 options:
-  -h, --help     show this help message and exit
-  -w, --write    Write the changes to file instead of printing the diffs to stdout
-  -v, --version  Show version number and exit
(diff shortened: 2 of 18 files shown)


tests/data/config/exclude_match/pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.pydocstringformatter]
+exclude = "**/test_package/**"
```
Pierre-Sassoulas, line 2: There's no newline at end of file there, maybe we could make the pre-commit exclude for newline and white space wider ?

Pierre-Sassoulas (approved):
Nice feature !

tests/data/config/exclude_match/pyproject.toml
DanielNoord: We could although I think stripping white spaces at the end of an docstring would also be a nice feature for us, so we'll need to then turn it off for the sub directory again?

Pierre-Sassoulas: I think we should not step on the toes of pre-commit's ``trailing-whitespace`` and ``end-of-file-fixer``. But we can deactivate specifically test files for it like in pylint if we want to handle whitespaces. (https://github.com/PyCQA/pylint/blob/main/.pre-commit-config.yaml#L9)

DanielNoord: Done with https://github.com/DanielNoord/pydocstringformatter/commit/71534deed2de0cd7254da5fc6699471a3a462d4b

Issue #25, opened 2022-01-07
Splitting summary and following line


Pierre-Sassoulas:
With the following code:
```python
    def other_inner():
        """A docstring with a sentence ending with dot. And another one
        continuing later."""
```
The result would be:
```python
    def other_inner():
        """A docstring with a sentence ending with dot.

        And another one continuing later."""
```

Could be named `SplitSummaryAndDocstringFormatter` and splits on periods?

See original discussion : https://github.com/DanielNoord/pydocstringformatter/pull/23#discussion_r779909114

PR #32, opened 2022-01-09
Add a testutil class to be able to test formatters easily

Pierre-Sassoulas:
This will permit to reuse the code to assert that a formatter does something if activated or nothing if deactivated more easily.


Pierre-Sassoulas:
Permit to do #25 's test with:
```python
def test_optional_formatters_argument(
    capsys: pytest.CaptureFixture[str],
    tmp_path: Path,
) -> None:
    """Test that an optional formatter is correctly turned on and off with arguments."""
    with FormatterAssert(
        '"""Summary. Body."""', [SplitSummaryAndDocstringFormatter()], capsys, tmp_path
    ) as asserter:
        asserter.assert_format_when_activated()
        asserter.assert_no_change_when_deactivated()
```

DanielNoord:
Just two general comments:
1. I think we should make everything in `testutils` private with a leading underscore for now. (As long as we don't support plugins nobody would need it anyway).
2. Is there any use case for those properties outside of testing? If not I think it would be better to make them into utility functions within the testutils package. Not really a need to add functions to the codebase that are only necessary for testing imo.

DanielNoord:
@Pierre-Sassoulas Have you had time to look at this?

Pierre-Sassoulas:
No sorry I have a pretty busy week, it's on my radar though.

> if not I think it would be better to make them into utility functions within the testutils package. 

Did you mean a function inside `tests` ?

DanielNoord:
> No sorry I have a pretty busy week, it's on my radar though.

No rush!

> > if not I think it would be better to make them into utility functions within the testutils package.
> 
> Did you mean a function inside `tests` ?

I think in `testutils`. Like: if we don't need those Class properties for the normal functionality of the program I think we shouldn't put them on the class (as they impact runtime of the program). So everything that is only necessary to test the program should be in `testutils` so it only impacts the running of the test suite instead of the actual runtime of the program.

Pierre-Sassoulas:
Allright I understand.

>  Like: if we don't need those Class properties for the normal functionality of the program 

We could refactor the argparsing to use them (I did not do that yet.) Maybe we could do the refactor first ? what do you think ?

DanielNoord:
> Allright I understand.
> 
> > Like: if we don't need those Class properties for the normal functionality of the program
> 
> We could refactor the argparsing to use them (I did not do that yet.) Maybe we could do the refactor first ? what do you think ?

Yeah makes sense. I'll do a final check of this PR > update the private names > merge and do a follow-up PR to `argparse`.

DanielNoord:
Follow up in #37.

Pierre-Sassoulas:
Thank you for finalizing this for me 😅

Issue #36, opened 2022-02-01
Make pushing summary to first line optional


KelSolaar:
Hi,

It is common with Sphinx to write module titles that can be used with `sphinx-apidoc` and `autodoc`, for example, the [Sphinx Gallery plugin](https://sphinx-gallery.github.io/stable/syntax.html) uses them.

We also do use them but unfortunately, the title gets pushed back on the first line:

```python
"""Spectral Uniformity
===================

Defines the objects to compute the *spectral uniformity*
(or *spectral flatness*) of spectral distributions.

References
----------
-   :cite:`David2015a` : David, A., Fini, P. T., Houser, K. W., Ohno, Y.,
    Royer, M. P., Smet, K. A. G., Wei, M., & Whitehead, L. (2015). Development
    of the IES method for evaluating the color rendition of light sources.
    Optics Express, 23(12), 15888. doi:10.1364/OE.23.015888
"""
```

I was wondering if there is a case to bypass the formatting of the two first lines.

Cheers,

Thomas

DanielNoord:
I'm wondering if we should make pushing the summary of multi-strings back to the first line optional. It's not actually "demanded" by the PEP and just something I added as it was easy to implement.
I think it makes sense to make it optional.

KelSolaar:
That would make sense to me! I actually managed to find the related conversation on Pydocstyle I was looking for before opening this issue: https://github.com/PyCQA/pydocstyle/issues/418

DanielNoord:
@KelSolaar Just to get this right myself:
1) Do titles still require a period after their summary? I think not enforcing the period runs counter to what the PEPs say. I'd like to stick to them to avoid too much discussion about semantics/details (that's what this tool is aiming to avoid anyway).
2) Can the second line only be made up of `====` or are there other patterns we should allow?

KelSolaar:
Hi,

1. I think that the period would not be needed in that case.
2. RST does not enforce usage of `equal` to mark the title header, it uses whatever symbol comes first, then the second and so forth to define the hierarchy. Put another way:

```reStructuredText
My Title
========

My Section
----------

My Sub-Section
^^^^^^^^^^^^^^
```

Should be equivalent to

```reStructuredText
My Title
********

My Section
^^^^^^^^^^

My Sub-Section
==============
```

PR #40, opened 2022-02-07
Special case Sphinx-style titles for ``FinalPeriodFormatter``

DanielNoord:
First step towards #36.

This already works with the current code, but now when we change anything in `FinalPeriodFormatter` we make sure that this keeps working.
Second step would be to recognise Sphinx-style titles in ``BeginningQuotesFormatter``.

### DIFF
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -44,7 +44,7 @@ class FinalPeriodFormatter(StringFormatter):
 
     name = "final-period"
 
-    def _treat_string(self, tokeninfo: tokenize.TokenInfo, indent_length: int) -> str:
+    def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
         """Add a period to the end of single-line docstrings and summaries."""
         # Handle single line docstrings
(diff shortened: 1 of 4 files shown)


pydocstringformatter/formatting/formatter.py
```diff
@@ -44,23 +44,24 @@ class FinalPeriodFormatter(StringFormatter):
 
     name = "final-period"
 
-    def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
+    def _treat_string(self, tokeninfo: tokenize.TokenInfo, indent_length: int) -> str:
```
DanielNoord, line 47:
```suggestion
    def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
```

Pierre-Sassoulas (approved):
Looks good, although I don't use this kind of formatting so I might be missing something. Should we ping the original reporter ?

DanielNoord:
Original reporter "hearted" the opening post so I think they are aware of this PR. Note that this doesn't really change anything, it's just a failsafe + tests in case of any future changes. The real necessary changes for #36 still need to be made.

PR #47, opened 2022-02-10
Add ``QuotesTypeFormatter``

### LINKED ISSUE #45: Create invalid Python code when a function declaration is followed by a normal string
Trying to update ``pylint`` the following code creates invalid Python:

```python
 def __init__(self, mode):
        "init filter mo."
        __mode = 0
```

becomes
```
 def __init__(self, mode):
        "init filter mo."in
```

This would probably be solved by creating a `QuotesTypeFormatter`, which was on my to-do list anyway.
I'll create `QuotesTypeFormatter` for a quick `0.4.0` release and move all other issues to `0.5.0`.

DanielNoord:
Closes #45

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     rev: v4.1.0
     hooks:
       - id: trailing-whitespace
-        exclude: "tests/data/format/whitespace_stripper"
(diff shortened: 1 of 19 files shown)


Pierre-Sassoulas (approved):
👌 0.4.0 is going to be very nice !

PR #52, opened 2022-02-10
Add ':' and ';' to allowed punctuation

Pierre-Sassoulas:
This permit to stay sane for this:
```diff
-        """Check the name of first argument, expect:
+        """Check the name of first argument, expect:.
 
         * 'self' for a regular method
         * 'cls' for a class method or a metaclass regular method (actually
```

### DIFF
--- a/tests/data/format/final_period/exotic_line_ending.py.out
+++ b/tests/data/format/final_period/exotic_line_ending.py.out
@@ -1,0 +1,13 @@
+def function():
+    """Check the name of first argument, expect:
+
+      *'self' for a regular method
+      *'cls' for a class method or a metaclass regular method
+    """
+
+def function():
+    """This has peculiar line ending;
+
+    but it should not matter.
+
+    """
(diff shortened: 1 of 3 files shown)


tests/data/format/final_period/exotic_line_ending.py.out
```diff
@@ -0,0 +1,13 @@
...
+
+      *'self' for a regular method
+      *'cls' for a class method or a metaclass regular method
+    """
+
+def function():
+    """This has peculiar line ending;
+
+    but it should not matter.
+
```
Pierre-Sassoulas, line 12: I think this could be considered a minor issue, the spaces at the end were transformed to a newline instead of just removed.

DanielNoord: This sounds like a `Stripper` issue. We can't just strip all empty lines, but an empty line as last line should be strippable.

Can you open an issue?

DanielNoord (approved):

tests/data/format/final_period/exotic_line_ending.py.out
Pierre-Sassoulas: Opened #54

PR #53, opened 2022-02-10
Add ``--push-multiline-opening`` option

### LINKED ISSUE #36: Make pushing summary to first line optional

DanielNoord:
Closes #36.

Quite a lot to unpack here.
1) Lots of test changes. The `beginning_quote` folder has some duplicates now, but oh well. Doesn't hurt to test things twice I guess. The test suite still takes a minimal amount of time to run.
2) Formatters now have access to the `config` attribute. I could also pass this as a new argument but then some signatures would get 6 arguments.. That's a bit much.
3) Added the requested `--push-multiline-opening` option.
4) Fixed some issues that popped op because of this. For example, capitalisation of first letter that appear on the second line.

### DIFF
--- a/pydocstringformatter/configuration/arguments_manager.py
+++ b/pydocstringformatter/configuration/arguments_manager.py
@@ -76,6 +76,16 @@ class ArgumentsManager:
             ),
         )
 
+        self.configuration_group.add_argument(
+            "--summary-quotes-same-line",
+            action="store_true",
+            help=(
+                "Force the start of a multi-line docstring to be on the "
+                "same line as the opening quotes. Similar to how this is enforced "
+                "for single line docstrings."
+            ),
+        )
+
     def parse_options(
         self,
         argv: List[str],
--- a/pydocstringformatter/formatting/base.py
+++ b/pydocstringformatter/formatting/base.py
@@ -1,4 +1,5 @@
 import abc
+import argparse
 import re
 import tokenize
 from typing import Literal
@@ -8,6 +9,8 @@ class Formatter:
     """Base class for docstring formatter."""
 
     optional = False
+    config: argparse.Namespace
+    """Namespace object set when set_config_namespace is called."""
 
     @property
     @abc.abstractmethod
@@ -32,6 +35,10 @@ class Formatter:
     @abc.abstractmethod
     def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
         """Return a modified token."""
+
+    def set_config_namespace(self, config: argparse.Namespace) -> None:
+        """Set the config attribute for this formatter."""
+        self.config = config
 
 
 class StringFormatter(Formatter):
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -24,7 +24,7 @@ class CapitalizeFirstLetterFormatter(StringFormatter):
     """Capitalize the first letter of the docstring if appropriate."""
 
     name = "capitalize-first-letter"
-    first_letter_re = re.compile(r"""['"]{1,3}\s*(\w)""")
+    first_letter_re = re.compile(r"""['"]{1,3}\s*(\w)""", re.DOTALL)
 
     def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
         new_string = None
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -12,11 +12,26 @@ class BeginningQuotesFormatter(StringFormatter):
     """Fix the position of the opening quotes."""
 
     name = "beginning-quotes"
+    potential_single_line = re.compile(
+        r"""
+        ['"]{1,3}         # 3 opening quotes
+        \n\s*.+           # A line with any length of characters
+        \n\s*             # A line with only whitespace
+        ['"]{1,3}         # 3 ending quote
+    """,
+        re.X,
+    )
+    """Regex pattern to match against a potential single line docstring."""
 
     def _treat_string(self, tokeninfo: tokenize.TokenInfo, _: int) -> str:
         new_string = tokeninfo.string
         if new_string[3] == "\n":
-            new_string = re.sub(r"\n *", "", new_string, 1)
+            if (
+                new_string.count("\n") == 1  # Single line docstring
+                or self.config.summary_quotes_same_line  # Config for multi-line
+                or self.potential_single_line.match(new_string)  # Potential single line
+            ):
+                new_string = re.sub(r"\n *", "", new_string, 1)
         return new_string
 
 
--- a/README.md
+++ b/README.md
@@ -86,7 +86,8 @@ multi-line docstring
 
 **PEP 257: _The closing quotes are on the same line as the opening quotes_**
 
-For consistency this rule also gets applied to multi-line docstrings.
+This can be enforced on multi-line docstrings with the `--summary-quotes-same-line`
+option. This behaviour is turned off by default.
 
 ```python
 # Bad
@@ -101,11 +102,23 @@ My
 multi-line docstring
 """
 
+# With --summary-quotes-same-line
+"""
+My
+multi-line docstring
+"""
+
 # Good
 """My docstring"""
 
 """My docstring"""
 
+"""
+My
+multi-line docstring
+"""
+
+# With --summary-quotes-same-line
 """My
 multi-line docstring
 """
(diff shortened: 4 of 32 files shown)


DanielNoord:
@KelSolaar No idea if you're able to or if you got the time to do so, but with this change the tool should be more useable for you! Let me know if you got any feedback or if I missed an edge case!

KelSolaar:
Hey @DanielNoord,

I have just gave it a whirl and I think it works. Thank you!

DanielNoord:
@Pierre-Sassoulas Gentle ping 😄

pydocstringformatter/configuration/arguments_manager.py
```diff
@@ -69,6 +69,16 @@ def _register_arguments(self, version: str) -> None:
             ),
         )
 
+        self.configuration_group.add_argument(
+            "--push-multiline-opening",
```
Pierre-Sassoulas, line 73: This option name is hard to understand I had to read the description and the change in the functional tests.  I don't have a very clear option name to propose though. Maybe

```suggestion
            "--first-line-summary",
```


```suggestion
            "--multilines-with-summary",
```
?

pydocstringformatter/formatting/base.py
```diff
@@ -8,6 +9,8 @@ class Formatter:
     """Base class for docstring formatter."""
 
     optional = False
+    config: argparse.Namespace
+    """Namespace object set when set_config_namespace is called."""
```
Pierre-Sassoulas, line 13:
```suggestion
```
Maybe we could let child class create an attribute for the configuration of interest for them. Right now this configuration is a black box and you need to read a lot of code to see what's in it. If we have to create the attribute in the child class then you only have to read the child class.

pydocstringformatter/formatting/base.py
```diff
@@ -33,6 +36,10 @@ def deactivate_option(self) -> str:
     def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
         """Return a modified token."""
 
+    def set_config_namespace(self, config: argparse.Namespace) -> None:
+        """Set the config attribute for this formatter."""
+        self.config = config
+
```
Pierre-Sassoulas, line 42:
```suggestion
    def set_config_namespace(self, config: argparse.Namespace) -> None:
        """Set the config attribute for this formatter.
        
        This function need to be overridden if the formatter needs a configuration.
        """
        pass
```

pydocstringformatter/formatting/formatter.py
```diff
@@ -12,19 +12,26 @@ class BeginningQuotesFormatter(StringFormatter):
     """Fix the position of the opening quotes."""
 
     name = "beginning-quotes"
+    potential_single_line = re.compile(r"""['"]{1,3}\n\s*.+\n\s*['"]{1,3}""")
+    """Regex pattern to match against a potential single line docstring."""
```
Pierre-Sassoulas, line 17:
```suggestion
    def set_config_namespace(self, config: argparse.Namespace) -> None:
        self.push_multiline_opening =  config.push_multiline_opening
```

Pierre-Sassoulas (commented):
Thank you for pinging me I missed that.

pydocstringformatter/formatting/base.py
DanielNoord: Hmm, I don't immediately see what the benefit would be here.

We need a function in which the `config` attribute of the `Run` class is made accessible to the formatters. We can then set `config` as attribute on the formatter or only set the necessary attributes. But both would still require you to understand what's in the `Run.config` attribute. I could see a `set_specific_config_attributes` method (although I wonder how beneficial that would be), but I don't immediately see how not setting `self.config` here makes it easier to interact with the `Namespace` object later on. At some point you will need to know what you can and cannot extract form it.

pydocstringformatter/configuration/arguments_manager.py
DanielNoord: Hmm, agreed although those names also don't really fit the function I think.

It should be something with `multiline`, `quotes/position` and perhaps `summary`.

pydocstringformatter/formatting/base.py
Pierre-Sassoulas: The person that add the option has the entire context in mind when creating the option and setting it in their formatter here : https://github.com/DanielNoord/pydocstringformatter/pull/53#discussion_r807860798. By doing so they establish a clear list of relevant options for this formatter. The person reading the class to modify it later does not, but they can read ``set_config_namespace`` to get the full list of relevant option. The alternative for them is to have to read the full list of option in argparse and to ask oneself what option is relevant for the formatter. (Actually this is something I truly dislike about the way pylint handle option in checkers).

pydocstringformatter/configuration/arguments_manager.py
Pierre-Sassoulas: ``--summary-next-to-quote-for-multiline`` ? it's a little long but it could work.

DanielNoord: I went with ``--summary-quotes-same-line``. Would that be okay for you?

pydocstringformatter/formatting/base.py
DanielNoord: Hm, the downside of this is that the value of the option becomes detached from the `Namespace`. If for whatever reason the `Namespace` object changes (for example if we want to support different `Namespaces` for different directories) then we would need to update all formatters again.
I see how it would provide some clarity to somebody looking at the class, but on the other hand just searching for all occurrences of `self.config` will show you all relevant options as well.

pydocstringformatter/configuration/arguments_manager.py
Pierre-Sassoulas: Sounds fine, the long description will help anyway :)

pydocstringformatter/formatting/base.py
Pierre-Sassoulas: > If for whatever reason the Namespace object changes (for example if we want to support different Namespaces for different directories) then we would need to update all formatters again.

True. The question then become : How often are we going to change the namespace object vs reading the formatter's code ?

> but on the other hand just searching for all occurrences of self.config will show you all relevant options as well.

Fair enough, it can become overwhelming for pylint because checkers are huge sometime 2000+ lines. Not a problem here I guess. I'll let you decide :)

DanielNoord: > > If for whatever reason the Namespace object changes (for example if we want to support different Namespaces for different directories) then we would need to update all formatters again.
> 
> True. The question then become : How often are we going to change the namespace object vs reading the formatter's code ?

Not sure, although I think accessing new options in a formatter shouldn't happen too often. Until now all formatters have been relatively concise.

> > but on the other hand just searching for all occurrences of self.config will show you all relevant options as well.
> 
> Fair enough, it can become overwhelming for pylint because checkers are huge sometime 2000+ lines. Not a problem here I guess. I'll let you decide :)

Agreed and it's a good point to think about during the migration to `argparse`. Although I don't have an easy solution right away.

pydocstringformatter/formatting/formatter.py
```diff
@@ -12,19 +12,26 @@ class BeginningQuotesFormatter(StringFormatter):
     """Fix the position of the opening quotes."""
 
     name = "beginning-quotes"
+    potential_single_line = re.compile(r"""['"]{1,3}\n\s*.+\n\s*['"]{1,3}""")
```
Pierre-Sassoulas, line 15:
```suggestion
    potential_single_line = re.compile(r"""
        ['"]{1,3} # 3 opening quotes
        \n\s*.+\n\s* # ...
        ['"]{1,3} # 3 ending quote
    """)
```
Would something like this be helpful ? It's hard to maintain regex but maybe this one is simple enough.

Pierre-Sassoulas (approved):
LGTM, maybe we can document the regex a little better.

pydocstringformatter/formatting/formatter.py
DanielNoord:
```suggestion
    potential_single_line = re.compile(r"""
        ['"]{1,3}         # 3 opening quotes
        \n\s*.+           # A line with any length of characters
        \n\s*              # A line with only whitespace
        ['"]{1,3}         # 3 ending quote
    """)
```

DanielNoord:
Going to reorder the commits and then merge. Thanks for the suggestions @Pierre-Sassoulas

PR #58, opened 2022-02-15
Add a usage page to the documentation

DanielNoord:
Let's see how this works.

Edit: Left to do:
- Make sure the documentation builds within the `pre-commit`. Current challenge is forcing `pre-commit` to install `-e .` before running the `docs` makefile. That is necessary to build the most recent version of the documentation.

### DIFF
--- a/docs/_ext/link_fixer.py
+++ b/docs/_ext/link_fixer.py
@@ -1,0 +1,63 @@
+import re
+from typing import Any, Dict
+
+from docutils import nodes
+from sphinx import addnodes, application
+from sphinx.transforms import SphinxTransform
+
+DOCUMENTATION_LINK = "https://pydocstringformatter.readthedocs.io/en/latest/"
+
+
+class LinkTransformer(SphinxTransform):
+    """Transformer to make all internal links actually internal.
+
+    For example:
+    Transforms https://pydocstringformatter.readthedocs.io/en/latest/usage.html
+    into a link to docs/usage.rst.
+    """
+
+    # Set priority very low so that we run after everything else has been done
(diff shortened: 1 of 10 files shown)


docs/_ext/link_fixer.py
```diff
@@ -0,0 +1,61 @@
...
+    # Set priority very low so that we run after everything else has been done
+    default_priority = 1000
+
+    def apply(self, **_: Dict[str, Any]) -> None:
+        """Apply the transformation."""
+        for node in self.document.traverse(nodes.reference):
+            if (
+                "refuri" in node
+                and node["refuri"].startswith(
+                    "https://pydocstringformatter.readthedocs.io"
```
Pierre-Sassoulas, line 25: This should be a constant imo. Well we can search and replace it in this little script, so it's not that important.

Pierre-Sassoulas (approved):
LGTM

DanielNoord:
**Needs to be rebased due to commit order**

@Pierre-Sassoulas I made it a variable.

See https://github.com/DanielNoord/pydocstringformatter/runs/5253162943?check_suite_focus=true for what the CI check is doing.
Basically, if something has been changed that should be reflected in the docs and the docs haven't been rebuilt locally this will warn us.
At some point we might want to make this do an auto-commit to actually rebuild the documentation, but this will be good enough for now.

Pierre-Sassoulas (approved):
👌

DanielNoord:
Just tested and this works. Link on `readthedocs` links correctly and the one on the GitHub README as well. That's a pretty nice accomplishment (if I say so myself 😄).

PR #62, opened 2022-02-25
Backport ``BooleanOptionalAction`` and use it for formatters

### LINKED ISSUE #61: Backport ``BooleanOptionalAction`` from ``argparse``
See https://stackoverflow.com/questions/9234258/in-python-argparse-is-it-possible-to-have-paired-no-something-something-arg.

Shouldn't be too difficult.

DanielNoord:
Schermafbeelding 2022-02-25 om 18 03 17

New output. Looks a lot cleaner in my opinion.

Made some slight changes to the class, but in general it is the same as the original implementation.

Closes #61.


Pierre-Sassoulas (approved):
The code is pretty clean, the new output too ! Great job !

PR #69, opened 2022-03-13
Improve ``SplitSummaryAndDocstringFormatter``

DanielNoord:
I have tested this (partially) on `pylint` and with `max-summary-lines=2` I think we can turn this on for `pylint`.

I'd like to do that first and then release `0.6.0` with the wrapping of lines. That should make the diff somewhat more manageable.

### DIFF
--- a/tests/data/format/summary_splitter/dots.py.out
+++ b/tests/data/format/summary_splitter/dots.py.out
@@ -1,0 +1,16 @@
+class MyClass:
+    """We shouldn't split sys.path."""
+
+
+def func():
+    """We should not add extra lines after the dot
+    to it.
+    """
+
+
+def func():
+    """We should not split for e.g. here."""
+
+
+def func():
+    """We should not split for i.e. here."""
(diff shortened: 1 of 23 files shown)


tests/data/format/summary_splitter/dots.py.out
```diff
@@ -0,0 +1,16 @@
+class MyClass:
+    """We shouldn't split sys.path."""
+
+
+def func():
+    """We should not add extra lines after the dot
+    to it.
+    """
+
```
Pierre-Sassoulas, line 9:
```suggestion
def func():
    """We should not add line after i.e., sys.path, e.g., etc. but etc. in particular 
    
    is harder right ?
    """
```
Just wondering what happens here 😄 I'm sure it's not a very important use case.

Pierre-Sassoulas (commented):
Looks very good ! Looking forward to pydocstringformatter being upgradeable with pre-commit automatically 😄

tests/data/format/summary_splitter/dots.py.out
DanielNoord: If my regex skill haven't failed me this should pass 😄 

I'll check!

DanielNoord:
Added handling of `etc.`.

I wonder how this will interact with the automatic period adder for `pylint`. But that should be fixable in `0.6.0`.

Issue #67, opened 2022-03-13
Create a ``SummaryFormatter`` base class


DanielNoord:
A class that implements `_treat_summary` and only returns a new string that is a summary which then gets concatenated with the description in the base class. Would avoid some code duplication.

PR #72, opened 2022-03-14
Create ``SummaryFormatter``

DanielNoord:
First step towards #67, but not fully there yet.

I'd like it so the checks on L167 and L179 are no longer necessary. Basically the `_treat_string` of `SummaryFormatter` should trim those last piece of indentation before passing the summary a long.

For now, it is good enough though.


Pierre-Sassoulas (approved):
Nice refactor !

PR #77, opened 2022-03-16
Handle multi-line summaries in ``FinalPeriodFormatter``

### LINKED ISSUE #26: Handle multi-line docstrings without a one line summary in ``FinalPeriodFormatter``
Currently we do not do anything whenever the summary of a multi-line docstring is longer than one line.

We have multiple options, also dependent on whether we are going to enforce single line summaries for multi-line docstrings:

1. Add a period at the end of every first line irrespective of whether the "sentence" continues. This would be a very harsh way to enforce single-line summaries. I don't like it.
2. Add a period at the end of the summary if it stops at the second (or third) line
3. Leave as is and leave it up to another formatter to decide whether the summary should be one line.

The difficulty with the 3rd option is that if we implement such a checker users can still disable it so we can't assume it has been run. We should think about that when choosing.

Relevant lines:
https://github.com/DanielNoord/pydocstringformatter/blob/0c6691b08399ea20ed132791ba0ac5ab9004d162/pydocstringformatter/formatting/formatter.py#L64-L68

DanielNoord:
Closes #26.

`SummaryFormatter` makes this a lot easier. I did a quick test against `pylint` and didn't find any immediate regressions.

### DIFF
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -77,43 +77,23 @@ class ClosingQuotesFormatter(StringFormatter):
         return new_string
 
 
-class FinalPeriodFormatter(StringAndQuotesFormatter):
+class FinalPeriodFormatter(SummaryFormatter):
     """Add a period to the end of single line docstrings and summaries."""
 
     name = "final-period"
     END_OF_SENTENCE_PUNCTUATION = {".", "?", "!", "‽", ":", ";"}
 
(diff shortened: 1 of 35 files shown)


DanielNoord:
@Pierre-Sassoulas lots of clean up to do, but look what I hacked together 😄

Pierre-Sassoulas:
Damn that github primer action is SO good. Definitely thinking about what can be done with pylint's primers...

PR #78, opened 2022-03-17
Add primer tests

DanielNoord:
TODO List:

- [x] Create branch with changes in PR and branch with `merge-base` (see mypy)
- [x] Run pydocstringformatter with `merge-base` branch in write mode
- [x] Checkout to changes branch
- [x] Do a non-write branch to get diff
- [x] Prettify diff
- [x] Probably tons more...

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -42,3 +42,25 @@ To only run a specific test from that directory, for example
 .. code-block:: shell
 
   pytest -k multi_line-class_docstring
+
+
+Primer
+-------
+
+To check if any changes create any unforeseen regressions all pull requests are tested
+with our primer. This process is heavily inspired on other open source projects and our
+own primer is based on the one used by `mypy <http://mypy-lang.org>`_.
+
+You can also run the primer locally with minimal set-up. First you will need to make
+a duplicate of your ``pydocstringformatter`` directory in a new ``./program_to_test``
+directory. This is required so that the primer can check out multiple versions of the
+program and compare the difference.
+
+The next step is to follow the bash script in the ``Run primer`` step in the ``primer``
+workflow file, found at ``.github/workflows/primer.yaml``. This should allow you to run
+all necessary steps locally.
+
+The final output of the primer run can be found in ``.pydocstringformatter_primer_tests/fulldiff.txt``
+
+New projects to run the primer over can be added to the ``pydocstringformatter/testutils/primer/packages.py``
+file.
--- a/.github/workflows/primer.yaml
+++ b/.github/workflows/primer.yaml
@@ -1,0 +1,83 @@
+# Most of this is inspired by the mypy primer
+
+name: Primer
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  primer-run:
+    name: Run
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out working version
+        uses: actions/checkout@v3.0.0
+        with:
+          fetch-depth: 2
+      - name: Check out changeable version
+        uses: actions/checkout@v3.0.0
+        with:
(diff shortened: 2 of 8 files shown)


DanielNoord:
@Pierre-Sassoulas Small achievement! It works. There are changes on `main` that would lead to changes to `pylint`, but the new version of the primer only shows the changes that would be introduced by this PR.

I'm going to do some code cleanup and add some annotations/comments but I actually think this is almost mergeable in an alpha form for the project.

_It also gave me some good insight into how to develop this in the future for pylint._

Pierre-Sassoulas:
> It also gave me some good insight into how to develop this in the future for pylint.

Yeap, having that in pylint will make the new checkers and every modification SO MUCH BETTER. I'm super hyped in anticipation.

DanielNoord:
> Yeap, having that in pylint will make the new checkers and every modification SO MUCH BETTER. I'm super hyped in anticipation.

I have written most of the code here to be somewhat copyable to `pylint` as well. One of the big issues is the speed: we need to run the program twice for every package we "prime" one. For `pydocstringformatter` this isn't an issue. For `pylint` this is...
I was thinking about maybe creating a "needs-primer" label so we can only schedule primers when necessary.. Hm, first `2.13` though 😄

Pierre-Sassoulas:
> need to run the program twice for every package we "prime" one.

Unless we save the baseline in the repository ? We'd also need to specify the commit for each primer repository. OR maybe it's possible to make a job manual (I did not check the doc of github actions) ?

DanielNoord:
> Unless we save the baseline in the repository ? We'd also need to specify the commit for each primer repository. OR maybe it's possible to make a job manual (I did not check the doc of github actions) ?

Yeah, I was thinking about that as well. But I foresee issues with fixing false-positives and comparing those etc. There is probably a reason why `mypy` doesn't save the baseline and runs twice for every primer test. They have more than enough knowledgeable contributors to do that optimisation if they saw a possibility/necessity to do so.

Yeah we can. We might also want to start with a smaller subset of repositories to do false positive checking for. I know that `home-assistant` have actually disabled `pylint` in some of their CI because it takes too long. So us running over them is perhaps not the best choice.

DanielNoord:
@Pierre-Sassoulas Not sure if you have the bandwidth to check this. I can merge as is, but I left it open to give you the opportunity to review (and potentially spot any errors that could be copied to `pylint` when I eventually copy this).

docs/development.rst
```diff
@@ -42,3 +42,25 @@ To only run a specific test from that directory, for example
...
+own primer is based on the one used by `mypy <http://mypy-lang.org>`_.
+
+You can also run the primer locally with minimal set-up. First you will need to make
+a duplicate of your ``pydocstringformatter`` directory in a new ``./program_to_test``
+directory. This is required so that the primer can check out multiple versions of the
+program and compare the difference.
+
+The next step is to follow the bash script in the ``Run primer`` step in the ``primer``
+workflow file, found at ``.github/workflows/primer.yaml``. This should allow you to run
+all necessary steps locally.
```
Pierre-Sassoulas, line 61: Maybe we could create a test with pytest for that ? It seem automatable even if not very convenient because you need to duplicate the git repository.

Pierre-Sassoulas (approved):
To be fair I skimmed most of the code but LGTM. It looks great actually.

docs/development.rst
DanielNoord: Yeah, I think it does not offer much: when would the test pass or fail?
At some point I could explore trying to create a script that does the same and handles cloning the repository, but for now I think this is fine.

Pierre-Sassoulas: Ha right it makes no sense as a test. A script to set it up is more like it.

PR #81, opened 2022-03-17
Add ``--exit-code`` option and add number of files to output

### LINKED ISSUE #4: Exit with correct exit codes
Currently the program does nothing with exit codes. This might be useful when it's used in combination with CIs.

DanielNoord:
Closes #4.

Also moved some stuff around and updated the output a little.

I kept the proposed exit codes as it just makes everything future proof.

For future reference:
0 - Nothing happened, all files are fine.
1 - Internal error
2 - Config parsing error
4, 8, 16 - Empty for future use
32 - A file was changed or could be changed
64, 128, 256 - Empty for future use

### DIFF
--- a/pydocstringformatter/run.py
+++ b/pydocstringformatter/run.py
@@ -33,25 +33,24 @@ class _Run:
 
         self._check_files(self.config.files)
 
+    # pylint: disable-next=inconsistent-return-statements
     def _check_files(self, files: List[str]) -> None:
         """Find all files and perform the formatting."""
         filepaths = utils._find_python_files(files, self.config.exclude)
 
         is_changed = self._format_files(filepaths)
 
-        if not is_changed:  # pylint: disable=consider-using-assignment-expr
-            if len(filepaths) > 1:
-                files_string = f"{len(filepaths)} files"
-            else:
-                files_string = "1 file"
+        if is_changed:  # pylint: disable=consider-using-assignment-expr
+            return utils._sys_exit(32, self.config.exit_code)
 
-            utils._print_to_console(
-                f"Nothing to do! All docstrings in {files_string} are correct 🎉\n",
-                self.config.quiet,
-            )
-            utils._sys_exit(0, self.config.exit_code)
-        else:
-            utils._sys_exit(32, self.config.exit_code)
+        files_string = f"{len(filepaths)} "
+        files_string += "files" if len(filepaths) != 1 else "file"
+        utils._print_to_console(
+            f"Nothing to do! All docstrings in {files_string} are correct 🎉\n",
+            self.config.quiet,
+        )
+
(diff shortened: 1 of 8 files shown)


pydocstringformatter/run.py
```diff
@@ -21,18 +21,37 @@ def __init__(self, argv: Union[List[str], None]) -> None:
...
+            else:
+                files_string = "1 file"
+
+            utils._print_to_console(
+                f"Nothing to do! All docstrings in {files_string} are correct 🎉\n",
+                self.config.quiet,
+            )
+            utils._sys_exit(0, self.config.exit_code)
+        else:
+            utils._sys_exit(32, self.config.exit_code)
```
Pierre-Sassoulas, line 54:
```suggestion
        if is_changed:  # pylint: disable=consider-using-assignment-expr
            utils._sys_exit(32, self.config.exit_code)
            return # We may not sys exit depending on options
        files_string = f"{len(filepaths)}"
        files_string += "files" if len(filepaths) > 1 else "file"
        utils._print_to_console(
            f"Nothing to do! All docstrings in {files_string} are correct 🎉\n",
            self.config.quiet,
        )
        utils._sys_exit(0, self.config.exit_code)
```
You know my favorite thing in the world is premature return I just had to say it 😄 Depending on what you think of displaying the name of a file if it's unique there the suggestion might need some tweaking.

pydocstringformatter/run.py
```diff
@@ -21,18 +21,37 @@ def __init__(self, argv: Union[List[str], None]) -> None:
...
-        self._format_files(filepaths)
+        filepaths = utils._find_python_files(files, self.config.exclude)
+
+        is_changed = self._format_files(filepaths)
+
+        if not is_changed:  # pylint: disable=consider-using-assignment-expr
+            if len(filepaths) > 1:
+                files_string = f"{len(filepaths)} files"
+            else:
+                files_string = "1 file"
```
Pierre-Sassoulas, line 46:
```suggestion
                files_string = f"'{filepaths[0].name}'"
```
Maybe a nice little bonus ?

Pierre-Sassoulas (approved):
Looks great ! I have two small suggestions but feel free to disregard.

pydocstringformatter/run.py
DanielNoord: Fails with `IndexError` in some test cases. I'll keep it in mind for a future enhancement.

pydocstringformatter/run.py
DanielNoord: I changed it around a little bit. You actually found a small bug: for `0 files` checked due to exclude patterns we still said that 1 file was correct.

I changed it (and fixed the tests).

Pierre-Sassoulas (approved):
👍

PR #94, opened 2022-04-28
Make ``SplitSummaryAndDocstringFormatter`` non-optional

### LINKED ISSUE #68: Make ``SplitSummaryAndDocstringFormatter`` non-optional
After working on it a bit more in `0.5.0` and adding the `--max-summary-lines` option I think it is time to make this non-optional.

DanielNoord:
Closes #68.

Best viewed commit per commit. Sadly had to update a lot of test outputs as they do not adhere to the formatting of this checker.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -9,8 +9,8 @@ Current usage of ``pydocstringformatter``:
                                 [--exit-code]
                                 [--max-summary-lines MAX_SUMMARY_LINES]
                                 [--summary-quotes-same-line]
-                                [--split-summary-body  --no-split-summary-body]
                                 [--strip-whitespaces  --no-strip-whitespaces]
+                                [--split-summary-body  --no-split-summary-body]
                                 [--linewrap-full-docstring  --no-linewrap-full-docstring]
                                 [--beginning-quotes  --no-beginning-quotes]
                                 [--closing-quotes  --no-closing-quotes]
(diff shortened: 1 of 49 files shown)


DanielNoord:
@Pierre-Sassoulas Note that the primer is a little trigger happy as we don't run with any commands. Since we have ``max-summary-lines=2`` and the default being 1, it seems like a lot will change. But the changes are minimal.

Pierre-Sassoulas:
For thing like :
```diff
-    """Get values listed in <columns> from <stats> and <old_stats>,
+    """Get values listed in <columns> from <stats> and <old_stats>,.
```
We could maybe replace the ``,`` by a ``.`` instead ?

DanielNoord:
> For thing like :
> 
> ```diff
> -    """Get values listed in <columns> from <stats> and <old_stats>,
> +    """Get values listed in <columns> from <stats> and <old_stats>,.
> ```
> 
> We could maybe replace the `,` by a `.` instead ?

I thought about this, but then decided it was probably best to let the user decide whether they want to break the sentence there or not. Just replacing `,` with `.` might not always work so some user-input will always be needed.

docs/usage.rst
```diff
@@ -50,6 +50,14 @@ Current usage of ``pydocstringformatter``:
...
                             docstring start, 2) docstring end and 3) end of line.
                             (default: True)
+      --split-summary-body, --no-split-summary-body
+                            Activate or deactivate split-summary-body: Split the
+                            summary and body of a docstring based on a period and
+                            max length. The maximum length of a summary can be set
+                            with the --max-summary-lines option. This formatter is
+                            currently optional as its considered somwehat
+                            opinionated and might require major refactoring for
+                            existing projects. (default: True)
```
Pierre-Sassoulas, line 60:
```suggestion
                            with the --max-summary-lines option. (default: True)
```

Pierre-Sassoulas (changes requested):

Issue #90, opened 2022-04-19
Add option to specify max line length


DanielNoord:
Necessary after #14.

PR #105, opened 2022-05-23
Max line length

mdczaplicki:
Small update to have max-line-length as a argument.

Sorry, but didn't have time to make tests pass. Some magic happens below :stuck_out_tongue: 

Closes https://github.com/DanielNoord/pydocstringformatter/issues/90

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -6,9 +6,9 @@ Current usage of ``pydocstringformatter``:
 .. code-block:: shell
 
     usage: pydocstringformatter [-h] [-w] [--quiet] [-v] [--exclude EXCLUDE]
-                                [--exit-code]
-                                [--max-summary-lines MAX_SUMMARY_LINES]
+                                [--exit-code] [--max-summary-lines int]
                                 [--summary-quotes-same-line]
+                                [--max-line-length int]
                                 [--strip-whitespaces  --no-strip-whitespaces]
                                 [--split-summary-body  --no-split-summary-body]
                                 [--linewrap-full-docstring  --no-linewrap-full-docstring]
@@ -35,13 +35,15 @@ Current usage of ``pydocstringformatter``:
       --exit-code           Turn on if the program should exit with bitwise exit
                             codes. 0 = No changes, 32 = Changed files or printed
                             diff.
-      --max-summary-lines MAX_SUMMARY_LINES
+      --max-summary-lines int
                             The maximum numbers of lines a summary can span. The
                             default value is 1.
       --summary-quotes-same-line
                             Force the start of a multi-line docstring to be on the
                             same line as the opening quotes. Similar to how this
                             is enforced for single line docstrings.
+      --max-line-length int
+                            Maximum line length of docstrings.
 
     default formatters:
       these formatters are turned on by default
--- a/pydocstringformatter/configuration/arguments_manager.py
+++ b/pydocstringformatter/configuration/arguments_manager.py
@@ -96,6 +96,7 @@ class ArgumentsManager:
                 "The maximum numbers of lines a summary can span. "
                 "The default value is 1."
             ),
+            metavar="int",
         )
 
         self.configuration_group.add_argument(
@@ -106,6 +107,15 @@ class ArgumentsManager:
                 "same line as the opening quotes. Similar to how this is enforced "
                 "for single line docstrings."
             ),
+        )
+
+        self.configuration_group.add_argument(
+            "--max-line-length",
+            action="store",
+            default=88,
+            type=int,
+            help="Maximum line length of docstrings.",
+            metavar="int",
         )
 
     def parse_options(
--- a/pydocstringformatter/formatting/base.py
+++ b/pydocstringformatter/formatting/base.py
@@ -145,7 +145,7 @@ class SummaryAndDescriptionFormatter(StringAndQuotesFormatter):
         # Remove opening quotes
         summary = summary[quotes_length:]
 
-        # Prefix is the new-line + idententation for summaries that
+        # Prefix is the new-line + indentation for summaries that
         # are not on the same line as the opening quotes
         prefix = ""
         if summary.startswith("\n"):
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -5,6 +5,8 @@ coveralls==3.3.1
 pytest==6.2.5
 pytest-cov==3.0.0
 gitpython>=3
+pre-commit==2.19.0
+
 
 # Requirements for docs building
 -r docs/requirements-doc.txt
--- a/tests/data/format/linewrap_summary/max_line_length_50.args
+++ b/tests/data/format/linewrap_summary/max_line_length_50.args
@@ -1,0 +1,2 @@
+--linewrap-full-docstring
+--max-line-length=50
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ options:
   -w, --write           Write the changes to file instead of printing the diffs to stdout.
   --quiet               Do not print any logging or status messages to stdout.
   -v, --version         Show version number and exit.
+  --max-line-length     The maximum docstring line length. Default set to 88.
 
 configuration:
   --exclude EXCLUDE     A comma separated list of glob patterns of file path names not to be formatted.
--- a/pydocstringformatter/formatting/formatter.py
+++ b/pydocstringformatter/formatting/formatter.py
@@ -73,7 +73,7 @@ class LineWrapperFormatter(SummaryFormatter):
     ) -> str:
         """Wrap the summary of a docstring."""
 
-        line_length = 88
+        line_length = self.config.max_line_length
 
         # Without a description we need to consider the length including closing quotes
         if not description_exists:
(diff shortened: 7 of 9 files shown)


.gitignore
```diff
@@ -2,3 +2,5 @@
 pydocstringformatter.egg-info
 dist/
 docs/_build
+venv/
+.idea/
```
DanielNoord, line 6: I'd prefer if you could add these to your own (global) `gitignore`. We don't really add any directories that are specific to any IDE or local environment as those should ideally be ignored by the user.

docs/usage.rst
```diff
@@ -17,6 +17,7 @@ Current usage of ``pydocstringformatter``:
                                 [--capitalize-first-letter  --no-capitalize-first-letter]
                                 [--final-period  --no-final-period]
                                 [--quotes-type  --no-quotes-type]
+                                [--max-line-length MAX_LINE_LENGTH]
```
DanielNoord, line 20: It seems the CI fails here. You could try:
```python
cd docs
make html
```

That should update the docs.

pydocstringformatter/configuration/arguments_manager.py
```diff
@@ -108,6 +108,13 @@ def _register_arguments(self, version: str) -> None:
             ),
         )
 
+        self.configuration_group.add_argument(
+            "--max-line-length",
+            action="store_const",
```
DanielNoord, line 113:
```suggestion
            action="store",
```

See `argparse` [documentation](https://docs.python.org/3/library/argparse.html#action). `store_const` actually stores one value `88` in this case and doesn't allow specifying another value. `store` should work for us!

pydocstringformatter/formatting/base.py
```diff
@@ -145,7 +145,7 @@ def _separate_summary_and_description(
         # Remove opening quotes
         summary = summary[quotes_length:]
 
-        # Prefix is the new-line + idententation for summaries that
+        # Prefix is the new-line + indentation for summaries that
```
DanielNoord, line 148: Pff, you found quite a bit of spelling errors 😄

requirements-test.txt
```diff
@@ -5,6 +5,8 @@ coveralls==3.3.1
 pytest==6.2.5
 pytest-cov==3.0.0
 gitpython>=3
+pre-commit==2.19.0
```
DanielNoord, line 8: 👍

tests/data/format/max_line_length/function_docstrings.args
```diff
@@ -0,0 +1 @@
+--max-line-length 50
```
DanielNoord, line 1: I'm not sure this works. We might need:
```suggestion
--max-line-length=50
```

tests/data/format/max_line_length/function_docstrings.py
```diff
@@ -0,0 +1,13 @@
+def func():
```
DanielNoord, line 1: Basically the test should have 3 cases:
1) One multi-line summary clearly above 50 characters. > Summary should be wrapped.
2) One multi-line summary with a description above 50 chars. > Summary should be wrapped, but description shouldn't as we don't support that.
3) One single-line docstring that is 49 chars without the final quotes. See `LinesChunk`.
If it is a single-line docstring we need to add the final quotes on the same line, so the line is not actually 49 but 52, so we should indeed wrap. That is what `LinesChunck` is testing.

DanielNoord (changes requested):
Thank you very much for taking this on @mdczaplicki. Very much appreciated!


I left some comments, let me know if you have any questions!

tests/data/format/max_line_length/function_docstrings.args
mdczaplicki: Oh, that's why :D

.gitignore
Pierre-Sassoulas: (You can define a global gitignore so you don't have to add this to all your gitignores) https://stackoverflow.com/a/7335487/2519059

pydocstringformatter/configuration/arguments_manager.py
mdczaplicki: Where do I set it to default 88 then?

DanielNoord:
```suggestion
            action="store",
            default=88,
            type=int,
```

See the example in the argparse docs:
```python
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('--foo', default=42)
>>> parser.parse_args(['--foo', '2'])
Namespace(foo='2')
>>> parser.parse_args([])
Namespace(foo=42)
```

tests/data/format/max_line_length/function_docstrings.py
mdczaplicki: OK. Updated them. But for some reason it seems that `50` is not picked up in this test.

DanielNoord: I think this is because the option doesn't use `type`. You could add a breakpoint somewhere after `line_length = self.config.max_line_length` but I think `line_length` will now be `"50"` instead of `50`. This can be fixed by using `type=int` in the `add_argument` call.

docs/usage.rst
mdczaplicki: Did that. Unfortunately pipelines still complain

tests/data/format/max_line_length/function_docstrings.py
mdczaplicki: It actually says that all docstrings are properly formatted. So it doesn't run

tests/data/format/linewrap_summary/max_line_length_50.args
```diff
@@ -0,0 +1,2 @@
+--linewrap-full-docstring
```
DanielNoord, line 1: @mdczaplicki I have pushed some changes. Turns out I was being very dumb.

Since line wrapping is still experimental it is not enabled by default. So for the test to run, it also needs this flag set.

pydocstringformatter/configuration/arguments_manager.py
```diff
@@ -108,6 +109,15 @@ def _register_arguments(self, version: str) -> None:
             ),
         )
 
+        self.configuration_group.add_argument(
+            "--max-line-length",
+            action="store",
+            default=88,
+            type=int,
+            help="Set the maximum line length of docstrings.",
```
DanielNoord, line 117:
```suggestion
            help="Maximum line length of docstrings.",
```

docs/usage.rst
```diff
@@ -35,13 +35,15 @@ Current usage of ``pydocstringformatter``:
...
-      --max-summary-lines MAX_SUMMARY_LINES
+      --max-summary-lines int
                             The maximum numbers of lines a summary can span. The
                             default value is 1.
       --summary-quotes-same-line
                             Force the start of a multi-line docstring to be on the
                             same line as the opening quotes. Similar to how this
                             is enforced for single line docstrings.
+      --max-line-length int
+                            Set the maximum line length of docstrings.
```
DanielNoord, line 46:
```suggestion
                            Maximum line length of docstrings.
```

DanielNoord (approved):
One final spelling change.

I'm also using this PR to test some changes to the primer. Sorry for the merge commit spam 😅

tests/data/format/linewrap_summary/max_line_length_50.args
mdczaplicki: No problem. Thanks for the help here :D

DanielNoord: No worries and thanks very much for doing the initial work. It's much easier as maintainer to build on a PR that's 90% done than starting from scratch.

Sorry for the spam you're getting right now. There is something wrong with one of our test runs but I can only test it on a PR opened by somebody that is not a contributor... I'll clean the commits in this PR up after I have fixed the issue and then merge this!

mdczaplicki: Tests pass locally. :tada: 

BTW. I believe there is a problem with `make html`. I was running it, but pipelines complained.

Also:
```
❯ cd docs
❯ pip install -e .
Obtaining file:///home/marek/PycharmProjects/pydocstringformatter/docs
ERROR: file:///home/marek/PycharmProjects/pydocstringformatter/docs does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.
❯ pip install -r requirements-doc.txt
Obtaining file:///home/marek/PycharmProjects/pydocstringformatter/docs (from -r requirements-doc.txt (line 1))
ERROR: file:///home/marek/PycharmProjects/pydocstringformatter/docs (from -r requirements-doc.txt (line 1)) does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.
```

However it works when I go 
```
cd ..
``` 
and then
```
pip install -r docs/requirements-doc.txt
```

DanielNoord: Yeah it should be:
```python
pip install -e .
cd docs
make html
```

This is due to how `readthedocs` install dependencies from root. So we need to install them from root as well...

DanielNoord:
Thanks once again @mdczaplicki, also for prompting me to final fix our CI 🎉

PR #117, opened 2022-06-13
fix: write file using existing newlines (fixes #115)

### LINKED ISSUE #115: On Windows, spurious ^M characters are added to Unix formatted files
Our team uses a `.gitattributes` file of
```
* -text  
```
This means files checked out on Windows will often contain only Unix line endings: just `^J`s (LF), and no `^M`s (CRs).

But when these files are reformatted using pydocstringformatter, the modified lines now end in `^M^J`, but only those lines.

Perhaps pydocstringformatter can sense the line ending characters of the lines it's modifying, and use them instead of introducing `^M`s where none were there previously?

I think this a better solution than adding a config option, but a config option would be nice too, to be explicit.

If there's interest, I can try to cobble a PR, that accomplishes this task.

rasa:
Here's my proposed solution. It should work in all cases, as it will default to the system default, if the file has different line endings (`file.newlines` returns a tuple).

I manually tested it with Unix, Dos, and Mac formatted files, and they all retained their line endings.

### DIFF
--- a/pydocstringformatter/run.py
+++ b/pydocstringformatter/run.py
@@ -64,6 +64,9 @@ class _Run:
                 raise utils.ParsingError(
                     f"Can't parse {os.path.relpath(filename)}. Is it valid Python code?"
                 ) from exc
+            # Record type of newlines so we can make sure to use
+            # the same later on.
+            newlines = file.newlines
 
         for index, tokeninfo in enumerate(tokens):
             new_tokeninfo = tokeninfo
@@ -85,7 +88,15 @@ class _Run:
                 filename_str = str(filename)
 
             if self.config.write:
-                with open(filename, "w", encoding="utf-8") as file:
+                if isinstance(newlines, tuple):
+                    newlines = newlines[0]
+                    print(
+                        "Found multiple newline variants in "
+                        f"{os.path.abspath(filename_str)}. "
+                        "Using variant that occurred first.",
+                        file=sys.stderr,
+                    )
+                with open(filename, "w", encoding="utf-8", newline=newlines) as file:
                     file.write(tokenize.untokenize(changed_tokens))
                     utils._print_to_console(
                         f"Formatted {filename_str} 📖\n", self.config.quiet
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/tests/data/format/newlines/* -text
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,8 +3,9 @@ repos:
     rev: v4.3.0
     hooks:
       - id: trailing-whitespace
-        exclude: "tests/data/format/whitespace_stripper|tests/data/format/quotes_type|tests/test_config.py"
+        exclude: "tests/data/format/whitespace_stripper|tests/data/format/quotes_type|tests/data/format/newlines|tests/test_config.py"
       - id: end-of-file-fixer
+        exclude: "tests/data/format/newlines"
       - id: check-yaml
       - id: check-toml
         exclude: &test-data "tests/data"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ enable = [
 
 [tool.pylint.BASIC]
 no-docstring-rgx = "__.*__"
+good-names = "f"
 
 [tool.pylint.TYPECHECK]
 generated-members = "argparse.Namespace"
--- a/tests/data/format/newlines/dos.py
+++ b/tests/data/format/newlines/dos.py
@@ -1,0 +1,7 @@
+var = 1
+"""A multi-line.
+docstring."""
+
+var = 1
+"""A single-line docstring.
+"""
--- a/tests/data/format/newlines/dos.py.out
+++ b/tests/data/format/newlines/dos.py.out
@@ -1,0 +1,8 @@
+var = 1
+"""A multi-line.
+
+docstring.
+"""
+
+var = 1
+"""A single-line docstring."""
--- a/tests/data/format/newlines/mac.py
+++ b/tests/data/format/newlines/mac.py
@@ -1,0 +1,1 @@
+var = 1"""A multi-line.docstring."""var = 1"""A single-line docstring."""
--- a/tests/data/format/newlines/mac.py.out
+++ b/tests/data/format/newlines/mac.py.out
@@ -1,0 +1,1 @@
+var = 1"""A multi-line.docstring."""var = 1"""A single-line docstring."""
--- a/tests/data/format/newlines/mixed.err
+++ b/tests/data/format/newlines/mixed.err
@@ -1,0 +1,1 @@
+Found multiple newline variants in {testfile}. Using variant that occurred first.
--- a/tests/data/format/newlines/mixed.py
+++ b/tests/data/format/newlines/mixed.py
@@ -1,0 +1,15 @@
+var = 1
+"""A multi-line.
+docstring."""
+
+var = 1
+"""A single-line docstring.
+"""
+
+var = 1
+"""A multi-line.
+docstring."""
+
+var = 1
+"""A single-line docstring.
+"""
--- a/tests/data/format/newlines/mixed.py.out
+++ b/tests/data/format/newlines/mixed.py.out
@@ -1,0 +1,17 @@
+var = 1
+"""A multi-line.
+
+docstring.
+"""
+
+var = 1
+"""A single-line docstring."""
+
+var = 1
+"""A multi-line.
+
+docstring.
+"""
+
+var = 1
+"""A single-line docstring."""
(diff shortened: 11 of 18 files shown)


DanielNoord (changes requested):
I don't think this fully solves the issue. If a repo contains files that are explicitly kept with `CRLF` and the tool is run on `Linux` in a `CI` it would still try to convert them based on the `os` newline.

Could you also add a test for this? I think you can copy one of the normal functional test but only need to make sure the newline is not `LF`. That should trigger a failure on the Windows or Linux tests.

rasa:
Yes, I will add a test. Do you want me to add MacOS to the test matrix? A simple example is at
https://github.com/psf/black/blob/6140f04fe17ae7c542a29c207c9a45f6200f4e2c/.github/workflows/test.yml
which has evolved into
https://github.com/psf/black/blob/HEAD/.github/workflows/test.yml

DanielNoord:
> Yes, I will add a test. Do you want me to add MacOS to the test matrix? A simple example is at
> https://github.com/psf/black/blob/6140f04fe17ae7c542a29c207c9a45f6200f4e2c/.github/workflows/test.yml
> which has evolved into
> https://github.com/psf/black/blob/HEAD/.github/workflows/test.yml

I think for this particular use case testing against Windows and Linux is enough, as they have different standard line endings and thus test the behaviour.

At some point I'll probably add macOS, but that's likely a task for another PR!

.gitignore
```diff
@@ -2,3 +2,4 @@
 pydocstringformatter.egg-info
 dist/
 docs/_build
+**/__pycache__/
```
DanielNoord, line 5:
```suggestion
```

This should probably be ignored in your global `.gitignore` file.

rasa:
So [this](https://github.com/DanielNoord/pydocstringformatter/blob/77b36a264afb6c03330cf52b5f4e36f8e0dd69a9/.pre-commit-config.yaml#L5) pre-commit hook is futzing my test files. I had committed these files in `tests/data/newlines` with the following newline formats:
```
dos.py       : dos
mac.py       : mac
mixed.py     : dos/unix
none.py      : none
unix.py      : unix
```
but the pre-commit hook rewrote them using these line endings:
```
dos.py       : dos
mac.py       : mac/unix
mixed.py     : dos/unix
none.py      : unix
unix.py      : unix
```
All the tests will still succeed, but breaks the tests, imo. Should I dynamically write these files out in the test's setup?

DanielNoord:
> So [this](https://github.com/DanielNoord/pydocstringformatter/blob/77b36a264afb6c03330cf52b5f4e36f8e0dd69a9/.pre-commit-config.yaml#L5) pre-commit hook is futzing my test files. I had committed these files in `tests/data/newlines` with the following newline formats:
> 
> ```
> dos.py       : dos
> mac.py       : mac
> mixed.py     : dos/unix
> none.py      : none
> unix.py      : unix
> ```
> 
> but the pre-commit hook rewrote them using these line endings:
> 
> ```
> dos.py       : dos
> mac.py       : mac/unix
> mixed.py     : dos/unix
> none.py      : unix
> unix.py      : unix
> ```
> 
> All the tests will still succeed, but this breaks the tests, imo. Should I dynamically write these files out in the test's setup?

You can add the directory with the tests for this specific issue into the ignore of the pre-commit hook. 
https://github.com/DanielNoord/pydocstringformatter/blob/77b36a264afb6c03330cf52b5f4e36f8e0dd69a9/.pre-commit-config.yaml#L6
It's a regular expresion so adding `|dir/name/file/` should work!

tests/data/newlines/dos.py
```diff
@@ -0,0 +1,6 @@
+var = """A multi-line
```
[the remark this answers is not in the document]
DanielNoord: Let me better explain myself.

By using the existing testing framework of input and `.out` files with a docstring that requires minor formatting (for example moving the quotes of a multi-line docstring to a new line we can fully check if we are inserting all new lines of the correct type.
Everything under https://github.com/DanielNoord/pydocstringformatter/tree/main/tests/data/format is ran automatically.
So `tests/data/format/newline/...` would be a good place to put these. 

Then copying and combining https://github.com/DanielNoord/pydocstringformatter/blob/main/tests/data/format/multi_line/module_docstring.py and https://github.com/DanielNoord/pydocstringformatter/blob/main/tests/data/format/ending_quote/module_docstring.py but changing the newlines would be a good fit as it is a very simple but clear change which would test all forms of new line insertions/removal.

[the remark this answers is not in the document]
rasa: Yikes! That sounds good, but perhaps you could implement? I'm a little lost, as I've only just started with this code base.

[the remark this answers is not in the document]
DanielNoord: Would it be okay for you if I set up the files like you currently did but then not do the line endings? I must confess I have no idea how to replicate those on my own Mac 😅 😄

rasa (commented):
@DanielNoord Are you requesting this section be changed? All the tests pass, so I'm not sure what changes are needed.

DanielNoord:
> @DanielNoord Are you requesting this section be changed? All the tests pass, so I'm not sure what changes are needed.

Sorry, this is probably turning out more difficult than you had imagined.

I have pushed a commit to your branch which includes correct tests for all 5 cases. You don't need to make any changes there. However, the `mixed` test case doesn't pass as the current solution is unable to determine the right newline. This means we probably need to rethink the solution.

rasa:
> However, the `mixed` test case doesn't pass as the current solution is unable to determine the right newline. This means we probably need to rethink the solution.

Thanks for the help. Your solution is much simpler.

Re the mixed use case, files with non-standard newlines (`\n\r`, `\r\n\n`, etc.) are either text files that got corrupted in some way (such as older versions of Perforce) or they're not UTF-8 text files at all, but some other encoding, or even binary files. So let's just delete the use case.

Of course, a better solution would be to completely ignore processing these files. That way, non-UTF-8 files would never get corrupted. Thoughts?

rasa:
Actually, I don't think this test framework is going to work for the "none" case, as the system will write the file using the system default newline character, so on Linux, `\n`, and on Windows, `\r\n`, and none.py.out has `\n`s hardcoded.

rasa:
Ok, it's a little hacky, but all tests pass on Windows now. I think we can put this to bed. Agreed?

DanielNoord:
> Ok, it's a little hacky, but all tests pass on Windows now. I think we can put this to bed. Agreed?

Hmm, yeah I'm not sure about the `mixed` and `none` examples. I think you might have been right before: they are so specific (and broken) that perhaps we shouldn't support them. Handling the different platforms by itself is already good enough for most users I think?

rasa:
> Handling the different platforms by itself is already good enough for most users I think?

I'd leave the none/mixed tests in, even though they are edge cases, because you can never have too many tests (at least those that run instantaneously), and they do work. If we wanted, we could even add more extreme edge cases, such as utf8-bom, utf16, and non-text files with `\r`s and `\n`s in them.

DanielNoord:
Thanks a lot @rasa!

I cleaned up the commit history and made some final changes. I'm going to make an immediate patch release with this included :)

pydocstringformatter/run.py
```diff
@@ -85,7 +88,15 @@ def _format_file(self, filename: Path) -> bool:
                 filename_str = str(filename)
 
             if self.config.write:
-                with open(filename, "w", encoding="utf-8") as file:
+                if isinstance(newlines, tuple):
+                    newlines = newlines[0]
```
rasa, line 92: @DanielNoord Your changes and comments look great. My only concern is if we process a utf-16, or non-text file, the file would most likely get corrupted. While I like the idea of automatically fixing text files that have non-standard newlines, we don't have any logic here to see if the file is really a text file, or not. Would checking for characters `\x00-\x08,\x0b-\x0c,\x0e-\x1f` tell us if the file is a binary file or not?

Perhaps I'm being overly cautious, but I think it's best to avoid corrupting files as much as possible, as the user may not even know the file is non-standard, or accidentally fed in the wrong files. Your thoughts?

DanielNoord: Although that's a valid point I wonder how often such issues will occur in the wild. Either way, I would recommend to use non-write mode (so leave the `-w`) for any first time user or anybody who is not using a source control program.
If you're using `git` it's not too bad that one file is corrupted as you'll be able to easily revert anyway.

I think the benefit of being careful for very particular and unlikely examples in this case doesn't really weigh up against the benefits of keeping this a bit a simpler and doing some auto-formatting.

_Perhaps we should add the stuff about using `-w` with source control in the README._

rasa: Thanks again for your help, and thoughtful changes. If it's ever decided to make it "binary aware", I confirmed that dos2unix [uses](https://sourceforge.net/p/dos2unix/dos2unix/ci/master/tree/dos2unix/common.c#l2034) the characters `\x00-\x08,\x0b,\x0e-\x1f` to determine if a file is binary.

PR #122, opened 2022-06-20
Make everything private

DanielNoord:


pydocstringformatter/_configuration/arguments_manager.py
```diff
@@ -2,13 +2,13 @@
...
-from pydocstringformatter.configuration import (
+from pydocstringformatter._configuration import (
     command_line_parsing,
     formatter_options,
     toml_parsing,
 )
-from pydocstringformatter.configuration.validators import VALIDATORS
-from pydocstringformatter.formatting.base import Formatter
+from pydocstringformatter._configuration.validators import VALIDATORS
+from pydocstringformatter._formatting.base import Formatter
```
Pierre-Sassoulas, line 11: Are you sure you don't want to make the Formatter public ? It's what make this extendable :)

DanielNoord: I want to revisit https://github.com/DanielNoord/pydocstringformatter/pull/29 when I'm satisfied with the tool.

There are some last things I'd like to include before a `1.0` release but since those aren't a priority and quite difficult to get right I don't think I'll get to them soon.

After that I want to look at the public API. It's probably easier to do that from an "all private" situation than a "all public" situation.

Pierre-Sassoulas: Yeah that's definitely better than the current pylint deprecation "situation" 😄

Issue #125, opened 2022-07-03
Feature Request: Support for Numpy, Google, and rst Docstrings


adam-grant-hendry:
Would it be possible to add support for docstring styles? e.g. Numpy, Google, and rst?

DanielNoord:
Yes! I'd very much like to support this but since I don't use these myself I'm not sure what is actually breaking right now.

If you could provide me with some sample docstrings and what you like to see I can certainly work on this!

adam-grant-hendry:
@DanielNoord Thank you for your support! Thank you so much for this wonderful package!

Don't get me wrong: nothing is breaking. Users must choose which format they wish to support.

- [Numpy style docstring example](https://numpydoc.readthedocs.io/en/latest/example.html#example)
- [Google style docstring example](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
- [Sphinx style docstring example](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html)

The [VSCode autoDocstring extension](https://github.com/NilsJPWerner/autoDocstring) is written in TypeScript, which I don't know very well so I'm not able to parse its syntax. The closest alternative I've seen of porting this to python is [python-autodocstring](https://github.com/r-zip/python-autodocstring), but it is still very much in its infancy.

I believe these could be implemented using `jinja2` templating, but there may be other ways. Not sure if you have to parse the AST or use a library like [libCST](https://libcst.readthedocs.io/en/latest/). I've not experimented with those before, but I know the [python-qt-tools](https://github.com/python-qt-tools) repos have used `libCST` to automate adding type hinting for `PyQt`/`PySide` modules. 

Currently, there are three limitations of `autoDocstring` that I would LOVE to have in `pydocstringformatter`:

- Cannot reanalyze docstrings on save. I submitted a feature request [here](https://github.com/NilsJPWerner/autoDocstring/issues/179) over a year ago with still no response.

- The extension cannot parse existing docstrings and determine what should be added or removed (it can only read the function/class signatures and add boilerplate strings). For example, `pydocstyle` will error if there is a signature mismatch between your function parameters and what's listed in your docstring. If `pydocstringformatter` could auto-detect the mismatch and then add this missing argument with a boilerplate string `param (type): _enter description here_` and remove non-existing parameters, this would be HUGE!

[...]

PR #132, opened 2022-07-21
ENH: Add formatters for numpydoc section ordering and name/type spacing

DWesl:
Related to #125; I got the bits I could think how to automate for the docstring style I actually use ([numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard))

Should probably extend the tests for more types of docstrings and more sections within each docstring.

Ideally the section test would have docstrings on
- [X] function
- [ ] class
- [ ] method
- [x] generator
- [x] module
- [x] constant

and would test the ordering of some subset of these sections:
- [X] Parameters
- [X] Returns
- [X] Raises
- [ ] Examples
- [x] Yields
- [x] References

Formatters for:
- [X] name-colon parameter spacing (`x : float` not `x:float` or `x: float`)
- [X] Section ordering
- [x] Section spacing (blank line between sections)
- [x] Length of line of hyphens after section header (should be same length as section header)

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -9,7 +9,7 @@ Current usage of ``pydocstringformatter``:
                                 [--exit-code] [--max-summary-lines int]
                                 [--summary-quotes-same-line]
                                 [--max-line-length int]
-                                [--style {pep257} [{pep257} ...]]
+                                [--style {pep257,numpydoc} [{pep257,numpydoc} ...]]
                                 [--strip-whitespaces  --no-strip-whitespaces]
                                 [--split-summary-body  --no-split-summary-body]
                                 [--linewrap-full-docstring  --no-linewrap-full-docstring]
@@ -45,7 +45,7 @@ Current usage of ``pydocstringformatter``:
                             is enforced for single line docstrings.
       --max-line-length int
                             Maximum line length of docstrings.
-      --style {pep257} [{pep257} ...]
+      --style {pep257,numpydoc} [{pep257,numpydoc} ...]
                             Docstring styles that are used in the project. Can be
                             more than one.
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -18,6 +18,10 @@ Current usage of ``pydocstringformatter``:
                                 [--capitalize-first-letter  --no-capitalize-first-letter]
                                 [--final-period  --no-final-period]
                                 [--quotes-type  --no-quotes-type]
+                                [--numpydoc-section-order  --no-numpydoc-section-order]
+                                [--numpydoc-name-type-spacing  --no-numpydoc-name-type-spacing]
+                                [--numpydoc-section-spacing  --no-numpydoc-section-spacing]
+                                [--numpydoc-section-hyphen-length  --no-numpydoc-section-hyphen-length]
                                 [files ...]
 
     positional arguments:
@@ -82,6 +86,22 @@ Current usage of ``pydocstringformatter``:
                             Activate or deactivate quotes-type: Change all opening
                             and closing quotes to be triple quotes. Styles:
                             default. (default: True)
+      --numpydoc-section-order, --no-numpydoc-section-order
+                            Activate or deactivate numpydoc-section-order: Change
+                            section order to match numpydoc guidelines. Styles:
+                            numpydoc. (default: True)
+      --numpydoc-name-type-spacing, --no-numpydoc-name-type-spacing
+                            Activate or deactivate numpydoc-name-type-spacing:
+                            Ensure proper spacing around the colon separating
+                            names from types. Styles: numpydoc. (default: True)
+      --numpydoc-section-spacing, --no-numpydoc-section-spacing
+                            Activate or deactivate numpydoc-section-spacing:
+                            Ensure proper spacing between sections. Styles:
+                            numpydoc. (default: True)
+      --numpydoc-section-hyphen-length, --no-numpydoc-section-hyphen-length
+                            Activate or deactivate numpydoc-section-hyphen-length:
+                            Ensure hyphens after section header lines are proper
+                            length. Styles: numpydoc. (default: True)
 
     optional formatters:
       these formatters are turned off by default
--- a/pydocstringformatter/_configuration/arguments_manager.py
+++ b/pydocstringformatter/_configuration/arguments_manager.py
@@ -123,7 +123,7 @@ class ArgumentsManager:
             action="extend",
             type=str,
             nargs="+",
-            choices=["pep257"],
+            choices=["pep257", "numpydoc"],
             help="Docstring styles that are used in the project. Can be more than one.",
         )
 
--- a/pydocstringformatter/_formatting/base.py
+++ b/pydocstringformatter/_formatting/base.py
@@ -5,7 +5,12 @@ import argparse
 import functools
 import re
 import tokenize
-from typing import Literal
+from collections import OrderedDict
+from collections.abc import Iterator
+from itertools import tee
+from typing import Literal, TypeVar
+
+_T = TypeVar("_T")
 
 
 class Formatter:
@@ -202,3 +207,92 @@ class SummaryFormatter(SummaryAndDescriptionFormatter):
 
     def treat_description(self, description: str, indent_length: int) -> str:
         return description
+
+
+def _pairwise(iterator: Iterator[_T]) -> Iterator[tuple[_T, _T]]:
+    """Create an iterator over pairs of successive elements."""
+    first, second = tee(iterator)
+    next(second)
+    return zip(first, second)
+
+
+class NumpydocSectionFormatter(StringAndQuotesFormatter, metaclass=abc.ABCMeta):
+    """Base class for formatters working on numpydoc sections."""
+
+    style = ["numpydoc"]
+
+    @abc.abstractmethod
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Process the individual numpydoc sections."""
+
+    def treat_string(
+        self,
+        tokeninfo: tokenize.TokenInfo,
+        indent_length: int,
+        quotes: str,
+        quotes_length: Literal[1, 3],
+    ) -> str:
+        """Split numpydoc sections, pass them for processing, then rejoin them."""
+        lines = tokeninfo.string[quotes_length:-quotes_length].split("\n")
+        # Handle the spaces before the closing quotes
+        last_line = lines[-1]
+        if lines[-1].isspace():
+            lines[-1] = ""
+
+        # Split sections
+        section_hyphen_lines = [
+            index
+            for index, line in enumerate(lines)
+            if "-" in line and all(char in " \t-" for char in line)
+        ]
+        section_starts = (
+            [0] + [index - 1 for index in section_hyphen_lines] + [len(lines)]
+        )
+        sections = OrderedDict(
+            [
+                (
+                    lines[curr_section_start].lstrip(),
+                    lines[curr_section_start:next_section_start],
+                )
+                for curr_section_start, next_section_start in _pairwise(
+                    iter(section_starts)
+                )
+            ]
+        )
+
+        if not section_hyphen_lines or section_hyphen_lines[0] > 1:
+            # The "Summary" section here includes the numpydoc
+            # summary, deprecation warning, and extended summary
+            # sections.  There's not an easy split for those the way
+            # there is for the other sections.
+            _, summary_section = sections.popitem(last=False)
+            sections["Summary"] = summary_section
+            sections.move_to_end("Summary", last=False)
+
+        # Process sections
+        new_sections = self.treat_sections(sections)
+
+        # Check that indent on first line of section didn't get weird
+        first_section = True
+        for section in new_sections.values():
+            if first_section:
+                section[0] = section[0].lstrip()
+                first_section = False
+            elif not section[0][0].isspace():
+                section[0] = f"{' ' * indent_length:s}{section[0]:s}"
+
+        # Rejoin sections
+        lines = [line for section in new_sections.values() for line in section]
+        # Ensure the last line puts the quotes in the right spot
+        if lines and lines[-1] == "":
+            if (last_line == "") or last_line.isspace():
+                # Try to preserve unindented closing quotes
+                lines[-1] = last_line
+            else:
+                # Quotes weren't on a different line before formatting
+                # but are now
+                lines[-1] = indent_length * " "
+        body = "\n".join(lines)
+        return f"{quotes:s}{body:s}{quotes:s}"
--- a/pydocstringformatter/_formatting/formatters_numpydoc.py
+++ b/pydocstringformatter/_formatting/formatters_numpydoc.py
@@ -1,0 +1,131 @@
+from __future__ import annotations
+
+from collections import OrderedDict
+
+from pydocstringformatter._formatting.base import NumpydocSectionFormatter
+
+
+class NumpydocSectionOrderingFormatter(NumpydocSectionFormatter):
+    """Change section order to match numpydoc guidelines."""
+
+    name = "numpydoc-section-order"
+
+    numpydoc_section_order = (
+        "Summary",
+        "Parameters",
+        "Attributes",
+        "Methods",
+        "Returns",
+        "Yields",
+        "Receives",
+        "Other Parameters",
+        "Raises",
+        "Warns",
+        "Warnings",
+        "See Also",
+        "Notes",
+        "References",
+        "Examples",
+    )
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Sort the numpydoc sections into the numpydoc order."""
+        for sec_name in reversed(self.numpydoc_section_order):
+            try:
+                sections.move_to_end(sec_name, last=False)
+            except KeyError:
+                pass
+        return sections
+
+
+class NumpydocNameColonTypeFormatter(NumpydocSectionFormatter):
+    """Ensure proper spacing around the colon separating names from types."""
+
+    name = "numpydoc-name-type-spacing"
+
+    numpydoc_sections_with_parameters = (
+        "Parameters",
+        "Attributes",
+        "Returns",
+        "Yields",
+        "Receives",
+        "Other Parameters",
+        "See Also",
+    )
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Ensure proper spacing around the colon separating names from types."""
+        for section_name, section_lines in sections.items():
+            if section_name in self.numpydoc_sections_with_parameters:
+                # Any section that gets here has a line of hyphens
+                initial_indent = section_lines[1].index("-")
+                for index, line in enumerate(section_lines):
+                    if (
+                        # There is content on this line (at least the initial indent)
+                        len(line) > initial_indent
+                        # and the first character after the indent for
+                        # the docstring is a name, not an additional
+                        # indent indicating a description rather than
+                        # a line with name and type
+                        and not line[initial_indent].isspace()
+                        # and there is a colon to separate the name
+                        # from the type (functions returning only one
+                        # thing don't put a name in their "Returns"
+                        # section)
+                        and ":" in line
+                    ):
+                        line_name, line_type = line.split(":", 1)
+                        if line_type:
+                            # Avoid adding trailing whitespace
+                            # Colon ending first line is suggested for long
+                            # "See Also" links
+                            section_lines[
+                                index
+                            ] = f"{line_name.rstrip():s} : {line_type.lstrip():s}"
+        return sections
+
+
+class NumpydocSectionSpacingFormatter(NumpydocSectionFormatter):
+    """Ensure proper spacing between sections."""
+
+    name = "numpydoc-section-spacing"
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Ensure proper spacing between sections."""
+        for section_lines in sections.values():
+            last_line = section_lines[-1]
+            if not (last_line == "" or last_line.isspace()):
+                section_lines.append("")
+        return sections
+
+
+class NumpydocSectionHyphenLengthFormatter(NumpydocSectionFormatter):
+    """Ensure hyphens after section header lines are proper length."""
+
+    name = "numpydoc-section-hyphen-length"
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Ensure section header lines are proper length."""
+        first_section = True
+        for section_name, section_lines in sections.items():
+            if section_name != "Summary":
+                # Skip the summary, deprecation warning and extended
+                # summary.  They have neither a section header nor the
+                # line of hyphens after it.
+                indent_length = section_lines[1].index("-")
+                section_lines[1] = " " * indent_length + "-" * len(section_name)
+                if first_section:
+                    # If the second line were not hyphens, the section name
+                    # would be summary.  This assumes triple quotes, but that
+                    # seems fine for a multi-line docstring.
+                    section_lines[1] = f"{section_lines[1]:s}---"
+            first_section = False
+        return sections
--- a/tests/data/format/numpydoc/numpydoc_header_line.py.out
+++ b/tests/data/format/numpydoc/numpydoc_header_line.py.out
@@ -1,0 +1,63 @@
+"""Example module for numpydoc docstring style.
+References
+----------
+NumPy docstring style guide:
+https://numpydoc.readthedocs.io/en/latest/format.html#documenting-modules"""
+import math
+
+EULER_NUMBER = math.e
+"""Euler's number.
+
+Not related to Euler's constant (sometimes called the Euler-Mascheroni
+constant.
+References
+----------
+E (mathematical constant)
+https://en.wikipedia.org/wiki/E_(mathematical_constant)
+Notes
+-----
+It is the limit of ``(1 + 1/n)**n`` as n approaches infinity, so it
+is used in the equation for continuously-compouned interest.
+
+It is also the sum of the reciprocals of the whole numbers starting
+with zero, which is related to some calculus-related properties
+mathemeticians find elegant.
+"""
+
+
+def sincos(theta):
+    """Returns
+    ----------
+    sin: float
+        the sine of theta
+    cos: float
+        the cosine of theta
+    Raises
+    ------
+    TypeError
+        If `theta` is not a float.
+    Parameters
+    ----------
+    theta: float
+        the angle at which to calculate the sine and cosine.
+    """
+    return math.sin(theta), math.cos(theta)
+
+
+def fibbonacci():
+    """Generate the Fibonacci sequence.
+
+    Each term is the sum of the two previous; conventionally starts
+    with two ones.
+    References
+    ----------
+    Fibonacci numbers
+    https://en.wikipedia.org/wiki/Fibonacci_number
+    Yields
+    ------
+    int"""
+    curr = 1
+    last = 0
+    while True:
+        yield curr
+        curr, last = curr + last, curr
--- a/tests/data/format/numpydoc/numpydoc_section_ordering.py.out
+++ b/tests/data/format/numpydoc/numpydoc_section_ordering.py.out
@@ -1,0 +1,63 @@
+"""Example module for numpydoc docstring style.
+References
+-----
+NumPy docstring style guide:
+https://numpydoc.readthedocs.io/en/latest/format.html#documenting-modules"""
+import math
+
+EULER_NUMBER = math.e
+"""Euler's number.
+
+Not related to Euler's constant (sometimes called the Euler-Mascheroni
+constant.
+Notes
+---
+It is the limit of ``(1 + 1/n)**n`` as n approaches infinity, so it
+is used in the equation for continuously-compouned interest.
+
+It is also the sum of the reciprocals of the whole numbers starting
+with zero, which is related to some calculus-related properties
+mathemeticians find elegant.
+
+References
+-----
+E (mathematical constant)
+https://en.wikipedia.org/wiki/E_(mathematical_constant)"""
+
+
+def sincos(theta):
+    """Parameters
+    -----
+    theta: float
+        the angle at which to calculate the sine and cosine.
+
+    Returns
+    ----
+    sin: float
+        the sine of theta
+    cos: float
+        the cosine of theta
+    Raises
+    ---
+    TypeError
+        If `theta` is not a float."""
+    return math.sin(theta), math.cos(theta)
+
+
+def fibbonacci():
+    """Generate the Fibonacci sequence.
+
+    Each term is the sum of the two previous; conventionally starts
+    with two ones.
+    Yields
+    ---
+    int
+    References
+    -----
+    Fibonacci numbers
+    https://en.wikipedia.org/wiki/Fibonacci_number"""
+    curr = 1
+    last = 0
+    while True:
+        yield curr
+        curr, last = curr + last, curr
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -241,6 +241,11 @@ class TestStyleOption:
         run = _Run([test_file, "--style", "pep257"])
         assert run.config.style == ["pep257"]
 
+    def test_style_numpydoc_only(self, test_file: str) -> None:
+        """Test that we can specify only a non-default style."""
+        run = _Run([test_file, "--style", "numpydoc"])
+        assert ["numpydoc"] == run.config.style
+
     def test_style_invalid_choice(self, capsys: pytest.CaptureFixture[str]) -> None:
         """Test that we correctly reject invalid styles."""
         with pytest.raises(SystemExit) as excinfo:
--- a/pydocstringformatter/_formatting/__init__.py
+++ b/pydocstringformatter/_formatting/__init__.py
@@ -13,6 +13,12 @@ from pydocstringformatter._formatting.formatters_default import (
     QuotesTypeFormatter,
     StripWhitespacesFormatter,
 )
+from pydocstringformatter._formatting.formatters_numpydoc import (
+    NumpydocNameColonTypeFormatter,
+    NumpydocSectionHyphenLengthFormatter,
+    NumpydocSectionOrderingFormatter,
+    NumpydocSectionSpacingFormatter,
+)
 from pydocstringformatter._formatting.formatters_pep257 import (
     SplitSummaryAndDocstringFormatter,
 )
@@ -31,4 +37,8 @@ FORMATTERS: list[Formatter] = [
     CapitalizeFirstLetterFormatter(),
     FinalPeriodFormatter(),
     QuotesTypeFormatter(),
+    NumpydocSectionOrderingFormatter(),
+    NumpydocNameColonTypeFormatter(),
+    NumpydocSectionSpacingFormatter(),
+    NumpydocSectionHyphenLengthFormatter(),
 ]
--- a/tests/data/format/numpydoc/numpydoc_header_line.args
+++ b/tests/data/format/numpydoc/numpydoc_header_line.args
@@ -1,0 +1,6 @@
+--style=numpydoc
+--no-numpydoc-name-type-spacing
+--no-numpydoc-section-order
+--no-numpydoc-section-spacing
+--no-final-period
+--no-closing-quotes
--- a/tests/data/format/numpydoc/numpydoc_header_line.py
+++ b/tests/data/format/numpydoc/numpydoc_header_line.py
@@ -1,0 +1,1 @@
+numpydoc_style.py
--- a/tests/data/format/numpydoc/numpydoc_parameter_spacing.args
+++ b/tests/data/format/numpydoc/numpydoc_parameter_spacing.args
@@ -1,0 +1,7 @@
+--style=numpydoc
+--numpydoc-name-type-spacing
+--no-numpydoc-section-order
+--no-numpydoc-section-spacing
+--no-numpydoc-section-hyphen-length
+--no-final-period
+--no-closing-quotes
--- a/tests/data/format/numpydoc/numpydoc_parameter_spacing.py
+++ b/tests/data/format/numpydoc/numpydoc_parameter_spacing.py
@@ -1,0 +1,1 @@
+numpydoc_style.py
--- a/tests/data/format/numpydoc/numpydoc_parameter_spacing.py.out
+++ b/tests/data/format/numpydoc/numpydoc_parameter_spacing.py.out
@@ -1,0 +1,63 @@
+"""Example module for numpydoc docstring style.
+References
+-----
+NumPy docstring style guide:
+https://numpydoc.readthedocs.io/en/latest/format.html#documenting-modules"""
+import math
+
+EULER_NUMBER = math.e
+"""Euler's number.
+
+Not related to Euler's constant (sometimes called the Euler-Mascheroni
+constant.
+References
+-----
+E (mathematical constant)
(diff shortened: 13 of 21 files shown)


pydocstringformatter/_formatting/base.py
```diff
@@ -196,3 +200,59 @@ def treat_summary(
...
+            for curr_section_start, next_section_start in _pairwise(
+                iter(section_starts)
+            )
+        ]
+        # Process sections
+        new_sections = self.treat_sections(sections)
+
+        # Rejoin sections
+        new_lines = [line for section in new_sections for line in section]
+        if len(last_line) > 0:
```
Pierre-Sassoulas, line 255:
```suggestion
        if last_line:
```
Because an empty string is falsey.

pydocstringformatter/_formatting/base.py
```diff
@@ -196,3 +200,59 @@ def treat_summary(
...
+            last_line = ""
+
+        section_heading_lines = [
+            index
+            for index, line in enumerate(lines)
+            if "-" in line and all(char in " \t-" for char in line)
+        ]
+        section_starts = [index - 1 for index in section_heading_lines]
+        section_starts.append(len(lines))
+        section_starts.insert(0, 0)
```
Pierre-Sassoulas, line 242:
```suggestion
        section_starts = [0] + [index - 1 for index in section_heading_lines] + [len(lines)]
```

Pierre-Sassoulas (commented):
Looks pretty good. I was wondering if we should detect that a docstring is in numpy style, but it can be handled in configuration. (Do not add the formatter if you don't have numpy style docstring)

DanielNoord (commented):
I don't have the time for a full review right now but this is highly appreciated!

That said, I would like to handle `numpy` docstrings behind a ``--style=`` flag. I think for future maintainability that will be much easier.
Since in reality projects can use different styles at the same time `--style ` should probably be an `append` type with the default being `["pep257"]`. For the tests in this PR we would then use ``--style=numpy``.

I should have probably documented this somewhere in the issue. Sorry about that!

We could also first create the PR that adds ``--style`` to keep the size of PRs limited. I might be able to do this myself but that might take 2/3 weeks.

@DWesl I have quickly hacked together a PR to add the ``--style`` flag as I had some spare time. See #138. After that it should be trivial to add ``"numpy"`` to the available choices for the option.

DWesl:
To add numpy to the options, yes, but actually having that option do something will take a bit more work.  I think I could arrange for `"numpydoc" in run.config.style` to imply both options I have here without terribly much trouble, but it sounded like you had a different idea for how the code should work.

DanielNoord:
Yeah I haven't thought this through completely but I was thinking of perhaps adding a `style` attribute to `Formatter` which is `"pep257"` by default but can also be `"numpy"` and then add formatters to the list of formatters to run based on the `config.style` attribute.

Does that sound like it would work?

DWesl:
The straightforward way to do that involves having the default for the formatter options depend on the `--style` value, which I don't know how to do with `argparse`

DWesl:
It turns out `BooleanOptionalAction` doesn't type-check the default, so I can use `None` as the default for checkers associated with a particular style, then loop through the formatter options after parsing the options and set any values that are still `None` (i.e., they were not set on the command line) based on the values in `self.namespace.style`.

DanielNoord:
@DWesl Just want to say this is on my radar. I'm on holiday currently but I expect to get to this PR somewhere this week! 😄

docs/usage.rst
```diff
@@ -18,12 +18,14 @@ Current usage of ``pydocstringformatter``:
...
                                 [--quotes-type  --no-quotes-type]
+                                [--numpydoc-section-order  --no-numpydoc-section-order]
+                                [--numpydoc-name-type-spacing  --no-numpydoc-name-type-spacing]
                                 [files ...]
 
     positional arguments:
       files                 The directory or files to format.
 
-    options:
+    optional arguments:
```
DanielNoord, line 28:
```suggestion
    options:
```

pydocstringformatter/_configuration/arguments_manager.py
```diff
@@ -24,8 +24,13 @@ def __init__(self, version: str, formatters: list[Formatter]) -> None:
 
         # First register all argument groups, then add arguments
         self.configuration_group = self.parser.add_argument_group("configuration")
-        self.default_formatters_group = self.parser.add_argument_group(
-            "default formatters", "these formatters are turned on by default"
+        self.pep257_default_formatters_group = self.parser.add_argument_group(
+            "PEP257 default formatters",
+            "these formatters are turned on by default and by --style=pep257",
```
DanielNoord, line 29:
```suggestion
            "These formatters are turned on by default and by --style=pep257",
```

pydocstringformatter/_configuration/arguments_manager.py
```diff
@@ -24,8 +24,13 @@ def __init__(self, version: str, formatters: list[Formatter]) -> None:
...
+        )
+        self.numpydoc_default_formatters_group = self.parser.add_argument_group(
+            "Numpydoc default formatters",
+            "these formatters are turned on by default with --style=numpydoc",
```
DanielNoord, line 33:
```suggestion
            "These formatters are turned on by default with --style=numpydoc",
```

pydocstringformatter/_configuration/arguments_manager.py
```diff
@@ -148,6 +154,14 @@ def parse_options(
         if self.namespace.style is None:
             self.namespace.style = ["pep257"]
 
+        for formatter in self.formatters:
```
DanielNoord, line 157: Why is this additional looping necessary?

pydocstringformatter/_configuration/formatter_options.py
```diff
@@ -25,5 +28,5 @@ def register_arguments_formatters(
             action=BooleanOptionalAction,
             dest=name,
             help=f"Activate or deactivate {name}: {formatter.__doc__}",
-            default=not formatter.optional,
+            default=False if formatter.optional else None,
```
DanielNoord, line 31: Why is this now `False`?

pydocstringformatter/_formatting/base.py
```diff
@@ -14,6 +18,7 @@ class Formatter:
     optional = False
     config: argparse.Namespace
     """Namespace object set when set_config_namespace is called."""
+    style = "pep257"
```
DanielNoord, line 21:
```suggestion
    style = "pep257"
    """Name of the docstring style that is associated with this formatter."""
```

DanielNoord (commented):
I did a first review of the code associated with the adding of the new style. The formatters themselves look relatively straightforward and I didn't see anything obvious just now.

I'm really happy you worked on this as this seems like a very nice addition. Let me know if there is anything I can do to help, I'd like to merge and release this ASAP 😄

pydocstringformatter/_configuration/formatter_options.py
Pierre-Sassoulas: There's something wrong here, default can only be falsey whatever the value given.

DWesl: - If `formatter.optional==True`, `default=False`, same as before.  
- If `formatter.optional==False`, then whether to apply the filter or not depends on the value of `--style` passed on the command line and is not known when setting defaults.  [As mentioned above](https://github.com/DanielNoord/pydocstringformatter/pull/132#issuecomment-1200207883), I can set the default to a sentinel value (here `None` for simplicity) and use that to include or exclude formatters for a style *only* if they were not explicitly given on the command-line.

pydocstringformatter/_configuration/arguments_manager.py
DWesl: So that [the default value for any given formatter](https://github.com/DanielNoord/pydocstringformatter/pull/132#issuecomment-1200207883) depends on whether `formatter.style in self.namespace.style`, the latter of which is not known when setting default values of the parameters, but only at runtime.

That is, so that the PEP257 formatters are on by default for `--style=pep257` and `--style=pep257 --style=numpydoc` but not `--style=numpydoc`, in a way that allows for `--style=pep257 --no-quotes-type`, `--style=numpydoc --no-name-type-spacing`, or `--style=numpydoc --final-period`.

I don't see another way to have all the formatter activation values be booleans at the end of `parse_options` and also have all the default values, set in `register_arguments_formatters` in `__init__`, be dependent on the value of `--style`, determined in `parse_command_line_arguments` and three lines up.

pydocstringformatter/_configuration/formatter_options.py
Pierre-Sassoulas: This part is not specific to the numpy style formatter, so I think this should be handled elsewhere. We probably lack a test when the formatter is not optional and default should be true.

pydocstringformatter/_formatting/base.py
```diff
@@ -202,3 +205,61 @@ def treat_summary(
...
+            for curr_section_start, next_section_start in _pairwise(
+                iter(section_starts)
+            )
+        ]
+        # Process sections
+        new_sections = self.treat_sections(sections)
+
+        # Rejoin sections
+        new_lines = [line for section in new_sections for line in section]
+        if len(last_line) > 0:
```
DanielNoord, line 262:
```suggestion
        if last_line != "":
```

right?

pydocstringformatter/_formatting/base.py
```diff
@@ -202,3 +205,61 @@ def treat_summary(
...
+            new_lines[-1] = last_line
```
DanielNoord, line 263: Should this be appended or overwritten?

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,85 @@
...
+        "Warnings",
+        "See Also",
+        "Notes",
+        "References",
+        "Examples",
+    )
+
+    def treat_sections(self, sections: list[list[str]]) -> list[list[str]]:
+        """Sort the numpydoc sections into the numpydoc order."""
+        # section_names = [sec_lines[0].strip() for sec_lines in sections]
```
DanielNoord, line 32: ?

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,85 @@
...
+from pydocstringformatter._formatting.base import NumpydocSectionFormatter
+
+
+class NumpydocSectionOrderingFormatter(NumpydocSectionFormatter):
+    """Change section order to match numpydoc guidelines."""
+
+    name = "numpydoc-section-order"
+
+    numpydoc_section_order = (
+        # I'm folding summary, deprecation warning, and extended
```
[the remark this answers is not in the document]
DanielNoord: I don't really understand this comment. Summary isn't at the top right?

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,85 @@
...
+            sections[1:],
+            key=lambda sec_lines: self.numpydoc_section_order.index(
+                sec_lines[0].strip()
+            ),
+        )
+        new_sections.insert(0, sections[0])
+        return new_sections
+
+
+class NumpydocParameterNameColonTypeFormatter(NumpydocSectionFormatter):
```
DanielNoord, line 44:
```suggestion
class NumpydocColonTypeFormatter(NumpydocSectionFormatter):
```

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,85 @@
...
+        "Receives",
+        "Other Parameters",
+        "See Also",
+    )
+
+    def treat_sections(self, sections: list[list[str]]) -> list[list[str]]:
+        """Ensure proper spacing around the colon separating names from types."""
+        new_sections = []
+        for section_lines in sections:
+            section_name = section_lines[0].lstrip()
```
DanielNoord, line 63: This is something we tend to do a lot? Should `sections` be an `OrderedDict`? With the name as key already?

tests/test_config.py
```diff
@@ -241,6 +241,11 @@ def test_style_pep257(self, test_file: str) -> None:
         run = _Run([test_file, "--style", "pep257"])
         assert run.config.style == ["pep257"]
 
+    def test_style_numpydoc_only(self, test_file: str) -> None:
```
DanielNoord, line 244: 👍

pydocstringformatter/_configuration/formatter_options.py
DWesl: > We probably lack a test when the formatter is not optional and default should be true.

Yeah, I was realizing that yesterday.  Should that be similar to the tests for `self.namespace.style`?

DanielNoord: I don't think that is necessary anymore.

In the current version optional checks also depend on the version.

So ``--numpydoc-name-type-spacing`` only works if you also provide ``--style=numpydoc``. That might be a bit counter intuitive, but I think of the `style` config as a way to tell `pydocstringformatter` which styles it should look out for and then apply any optional or non-optional formatters on them.

pydocstringformatter/_formatting/base.py
DWesl: The sections are *usually* separated by a blank line, but the last one needs spaces so the closing quotes are indented properly.  I should probably write a formatter for the first one and check which I need to do for the second (if it's a blank line, I replace it with the line of spaces; if not, I append the line of spaces).

DanielNoord: Can't you use `indent_length` for this? It is in the arguments to this function specifically for these situations.

pydocstringformatter/_formatting/formatters_numpydoc.py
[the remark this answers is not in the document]
DWesl: For the numpydoc style, I'm assuming sections are separated by a line with a section name and a line of hyphens.  This means the summary (first line of docstring), deprecation warning (just after that, with no separator), and extended summary (everything from there to the first section separator) are treated together rather than separately.  This means I can't push the deprecation warning to the correct place or ensure proper spacing between the deprecation warning and extended summary.  It was also a lot easier to think about when I was writing.

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: For the docstring:
https://numpydoc.readthedocs.io/en/latest/format.html#sections
For the commented-out prototype code;
I probably thought I would need the section names, then realized I didn't for this formatter (there was a while when I wasn't sure if there was a way to separate numpydoc autoformatting into multiple formatters).  For that:
```suggestion
```

I should maybe also add a formatter for that line of hyphens to make sure it's the same length as the section name.

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: What usually throws me off is the space between the parameter name and colon, not between the colon and the type (I write `x: float` instead of `x : float`; `sphinxcontrib.ext.napolean` accepts both; `numpydoc` accepts only the second).  Maybe 
```suggestion
class NumpydocNameColonTypeFormatter(NumpydocSectionFormatter):
```
since it also applies to the returned values?

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: That might work.  Alternately, look through the existing base classes to see if there's one to strip off the (common) indentation before processing and put it back after (that would also sidestep the `last_line` complexities above).

tests/test_config.py
DWesl: It would probably be useful to check that the formatter activation values are getting set properly.  Should that go in these tests or in separate tests?

pydocstringformatter/_formatting/base.py
DWesl:
```suggestion
            if new_lines[-1] and new_lines[-1].isspace():
                new_lines[-1] = last_line
            else:
                new_lines.append(last_line)
```

DWesl: > Can't you use `indent_length` for this? It is in the arguments to this function specifically for these situations.

I'd forgotten about that.  Yes.

pydocstringformatter/_formatting/formatters_numpydoc.py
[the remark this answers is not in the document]
DWesl: On a related note, what do you want `pydocstringformatter` to do if someone decides to include a `Preconditions` section (not in the standard) or `Prameters` (misspelled)?  I think the current sorting implementation produces an error, but I could probably change that to put all the weird ones at the end.

pydocstringformatter/_configuration/arguments_manager.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, version: str, formatters: list[Formatter]) -> None:
             "default formatters", "these formatters are turned on by default"
         )
         self.optional_formatters_group = self.parser.add_argument_group(
-            "optional formatters", "these formatters are turned off by default"
+            "options", "these formatters are turned off by default"
```
DanielNoord, line 31:
```suggestion
            "optional formatters", "these formatters are turned off by default"
```

pydocstringformatter/_formatting/formatters_numpydoc.py
DanielNoord: Yeah that works for me!

pydocstringformatter/_formatting/formatters_numpydoc.py
DanielNoord: I don't think there is currently.

tests/test_config.py
DanielNoord: I think we already have tests for those. But if you can come up with new ones they should probably be their own tests/functions.

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class NumpydocSectionOrderingFormatter(NumpydocSectionFormatter):
+    """Change section order to match numpydoc guidelines."""
+
+    name = "numpydoc-section-order"
+
+    numpydoc_section_order = (
+        # We are folding summary, deprecation warning, and extended
+        # summary into "before first heading"
```
DanielNoord, line 13:
```suggestion
        # We skip summary, deprecation warning, and extended
        # summary
```

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,84 @@
...
+        "See Also",
+        "Notes",
+        "References",
+        "Examples",
+    )
+
+    def treat_sections(self, sections: list[list[str]]) -> list[list[str]]:
+        """Sort the numpydoc sections into the numpydoc order."""
+        # assert all(name in self.numpydoc_section_order for name in section_names[1:])
+        new_sections = sorted(
```
DanielNoord, line 33:
```suggestion
        # We skip summary, deprecation warning, and extended
        # summary
        new_sections = sorted(
```

DanielNoord (changes requested):

tests/test_config.py
DWesl: With the new `--style` options?  Admittedly that's probably less of an issue with #145 than my original implementation.

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: Should I make a separate one or just keep that as part of the NumpydocSectionsFormatter or whatever I called it?  Or just make `sections` an `OrderedDict` (there's no `.sortkeys(key)` method on `OrderedDict`, which means I'll need a couple of passes through the sections.

tests/test_config.py
DanielNoord: Yeah, it should no longer depend on the `style` options`. If `FormatterA` has style `["numpydoc"]` it can be "turned on" but still be disabled when `style` does not include `numpydoc`.

pydocstringformatter/_formatting/formatters_numpydoc.py
DanielNoord: I was thinking of letting `NumpydocSectionFormatter` pass an `OrderedDict` to `treat_sections` in its `treat_string`. `NumpydocSectionOrderingFormatter` can then return an `OrderedDict` of a different order.

DanielNoord: Although, since `Python` `3.6` dicts are also order dependant. So we could just rely on that fact? I'm not sure, `OrderedDict` is probably better.

DWesl: I think it's [3.7](https://docs.python.org/3/library/collections.html#collections.OrderedDict) that the `dict` ordering became a Python thing rather than a CPython and PyPy implementation detail.  `OrderedDict` makes that intention explicit, which will probably help with maintenance.

docs/usage.rst
DanielNoord, line 28: Which interpreter version are you using locally?

pydocstringformatter/_formatting/base.py
```diff
@@ -202,3 +206,70 @@ def treat_summary(
...
+                )
+            ]
+        )
+        # Process sections
+        new_sections = self.treat_sections(sections)
+
+        # Rejoin sections
+        new_lines = [line for section in new_sections.values() for line in section]
+        # Ensure the last line puts the quotes in the right spot
+        # Enforces indented closing quotes on the last line
```
DanielNoord, line 269: This probably shouldn't be handled by this formatter. Can we store whether the closing quotes were already on a new line and do this according to that?

We like to keep each formatter responsible for a single thing to provide optimal customisation for users.

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,135 @@
...
+        "References",
+        "Examples",
+    )
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Sort the numpydoc sections into the numpydoc order."""
+        new_sections = OrderedDict([sections.popitem(last=False)])
+        new_sections.update(
```
DanielNoord, line 37: Can't we remove L36 and just return a sorted `OrderedDict`?

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,135 @@
...
+        "Other Parameters",
+        "See Also",
+    )
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Ensure proper spacing around the colon separating names from types."""
+        new_sections = OrderedDict([])
+        for section_name, section_lines in sections.items():
```
DanielNoord, line 71: Is there a particular why we don't replace lines in-place here? You do so in `NumpydocSectionSpacingFormatter`.

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,135 @@
...
+    name = "numpydoc-section-header-line-length"
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Ensure section header lines are proper length."""
+        new_sections = OrderedDict([])
+        first_section = True
+        for section_name, section_lines in sections.items():
+            if first_section:
```
DanielNoord, line 126: But what if the first section isn't a summary? Could you add a test for this?

DanielNoord (changes requested):
This PR is becoming quite large. If you think of any other formatters please add them in another PR.

I think this might be my last comments 😄

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,135 @@
...
+                [
+                    (sec_name, sections[sec_name])
+                    for sec_name in sorted(
+                        sections.keys(),
+                        key=self.numpydoc_section_order.index,
+                    )
+                ]
+            )
+        )
+        return new_sections
```
DWesl, line 48: Another option, using `OrderedDict` features and avoiding the `KeyError` on weird sections:
```suggestion
        new_sections = sections.copy()
        for sec_name in self.numpydoc_section_order:
            try:
                new_sections.move_to_end(sec_name)
            except KeyError:
                pass
        for sec_name in new_sections.keys()[1:]:
            if sec_name not in self.numpydoc_section_order:
                new_sections.move_to_end(sec_name)
        return new_sections
```

docs/usage.rst
DWesl: Cygwin CPython 3.9; should I be using 3.8 or 3.7 for this?

pydocstringformatter/_formatting/base.py
DWesl: So, new test for that.  Would it help if I made all the test input files symlinks of `numpydoc_style.py`, with only the changes for that formatter reflected in the corresponding `.py.out` file?

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: I don't see an `OrderedDict.sort` method [here](https://docs.python.org/3/library/collections.html#collections.OrderedDict).  and the implementation [here](https://github.com/python/cpython/blob/8b34e914bba2ccd6ae39609410db49d0beb19cb1/Lib/collections/__init__.py#L78-L329) looks like the ordering is kept by a linked list rather than a list I can sort.

To merge line 36 with 37-44, I would need a consistent name for the summary/deprecation warning/extended summary section, which currently uses the summary as the name.  I can look into that, as it would simplify other things.

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: If it starts with
```python
    """Parameters
    ----------
```
then there is no summary section; I think the current code would change this to
```python
    """
    Parameters
    ------------
```
To fix this further, I would need a consistent name for the summary/deprecation warning/extended summary section.

pydocstringformatter/_formatting/base.py
DanielNoord: That might work well in this case yeah!

pydocstringformatter/_formatting/formatters_numpydoc.py
DanielNoord: See above.

Although I think this change is good, it should probably also be its own formatter. But let's add that in a follow up PR.

pydocstringformatter/_formatting/formatters_numpydoc.py
DanielNoord: Too bad there isn't a `move_to_front`.

DWesl: There's `move_to_back(key, last=False)`, which should do the same thing.  My main reason for ordering it this way is now moot, so I should be able to iterate through `reversed(self.numpydoc_section_order)` without a problem.

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: I *think* I fixed this.  Should I change a test to make sure?

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: Trying to avoid side-effect, I think?  Since I'm not consistent, I can change this to just modify the lines in their original lists and pass that back.

DanielNoord: Let's do that!

pydocstringformatter/_formatting/formatters_numpydoc.py
DanielNoord: Yes, I think that would be good!

DWesl: "Works" now, although interaction with the section ordering formatter can produce strange results:
```python
    """    Parameters
    ----------
    ...

Returns
    -------
    ...
```
I think there's a formatter to strip the leading space; not sure about indenting `Returns` properly.

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: Now in-place

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: Initial section now named "Summary" unless the initial section is zero lines, in which case the initial section shares a name with the next section and gets disappeared by the `dict` constructor.

pydocstringformatter/_formatting/base.py
DWesl: Whole bunch of symlinks created and `.args` files expanded

docs/usage.rst
DWesl: Was hoping it might be because I did `pip install -r requirements-test.txt` instead of `pip install -U -r requirements-test.txt`, but adding the -U didn't seem to change anything

pydocstringformatter/_formatting/base.py
```diff
@@ -202,3 +206,91 @@ def treat_summary(
...
+        # Split sections
+        lines = tokeninfo.string[quotes_length:-quotes_length].split("\n")
+        # handle the spaces before the closing quotes
+        last_line = lines[-1]
+        if lines[-1].isspace():
+            lines[-1] = ""
+
+        # Everything before the first section header is in a single
+        # summary/deprecation warning/extended summary section.  This
+        # ends up called "Summary".
```
DanielNoord, line 246:
```suggestion
        # ends up being called "Summary".
```

pydocstringformatter/_formatting/base.py
```diff
@@ -202,3 +206,91 @@ def treat_summary(
...
+                    lines[curr_section_start:next_section_start],
+                )
+                for curr_section_start, next_section_start in _pairwise(
+                    iter(section_starts)
+                )
+            ]
+        )
+        if section_hyphen_lines and section_hyphen_lines[0] == 1:
+            # No summary/deprecation warning/extended summary section
+            section_starts.pop(0)
```
DanielNoord, line 269: Can `section_starts` be a `set` to avoid this?

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,115 @@
...
+        "Examples",
+    )
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Sort the numpydoc sections into the numpydoc order."""
+        for sec_name in reversed(self.numpydoc_section_order):
+            try:
+                sections.move_to_end(sec_name, last=False)
```
DanielNoord, line 37: 😍

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,115 @@
...
+        """Ensure proper spacing around the colon separating names from types."""
+        for section_name, section_lines in sections.items():
+            if section_name in self.numpydoc_sections_with_parameters:
+                # Any section that gets here has a line of hyphens
+                initial_indent = section_lines[1].index("-")
+                for index, line in enumerate(section_lines):
+                    if (
+                        ":" in line
+                        and len(line) > initial_indent
+                        and not line[initial_indent].isspace()
```
DanielNoord, line 70: What is this line checking for?

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,115 @@
...
+    """Ensure proper spacing between sections."""
+
+    name = "numpydoc-section-spacing"
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Ensure proper spacing between sections."""
+        new_sections = OrderedDict([])
+        for section_name, section_lines in sections.items():
```
DanielNoord, line 89: Shall we replace in-place here as well?

pydocstringformatter/_formatting/formatters_numpydoc.py
```diff
@@ -0,0 +1,115 @@
...
+    """Ensure hyphens after section header lines are proper length."""
+
+    name = "numpydoc-section-hyphen-length"
+
+    def treat_sections(
+        self, sections: OrderedDict[str, list[str]]
+    ) -> OrderedDict[str, list[str]]:
+        """Ensure section header lines are proper length."""
+        new_sections = OrderedDict([])
+        for section_name, section_lines in sections.items():
```
DanielNoord, line 107: In-place here as well?

tests/data/format/numpydoc/numpydoc_header_line.py.out
```diff
@@ -0,0 +1,63 @@
...
+
+It is also the sum of the reciprocals of the whole numbers starting
+with zero, which is related to some calculus-related properties
+mathemeticians find elegant.
+"""
+
+
+def sincos(theta):
+    """Returns
+    -------
```
DanielNoord, line 30: This doesn't seem correct? Shouldn't it be a little longer?

tests/data/format/numpydoc/numpydoc_section_ordering.py.out
```diff
@@ -0,0 +1,63 @@
...
+mathemeticians find elegant.
+
+References
+-----
+E (mathematical constant)
+https://en.wikipedia.org/wiki/E_(mathematical_constant)"""
+
+
+def sincos(theta):
+    """    Parameters
```
DanielNoord, line 29: This should probably be fixed.

tests/data/format/numpydoc/numpydoc_section_ordering.py.out
```diff
@@ -0,0 +1,63 @@
...
+    -----
+    theta: float
+        the angle at which to calculate the sine and cosine.
+
+Returns
```
DanielNoord, line 34: This as well.

DanielNoord (changes requested):
Lots of different things going on in this PR, but I feel like we're getting there 😄

pydocstringformatter/_formatting/base.py
DWesl: I need that list sorted and I'm not sure that `set`s are ordered the way `dict`s have been since 3.7 or so.  This is the last time that variable is used, and dictionary semantics means most other problems are already dealt with; would you prefer I negated the condition and only included the other branch?

pydocstringformatter/_formatting/formatters_numpydoc.py
DWesl: A parameter description might look like

```python
    """Summarize.

    Parameters
    ----------
    x : float
        The x-coordinate: the argument of the function

        New Paragraph: Usually in radians.

    y: float
        The y-coordinate: the abscissa of the plot

    Returns
    """
```
If the line is empty, or just the standard indent for the docstring, there's nothing to format.  
If the line is indented more than the standard, it's a description, and not relevant to this formatter.
If there is no colon, the name or type is missing, and there is also nothing for the formatter to do.

In other words, the lines relevant to this formatter must 
1. have content,
2. be indented flush with the heading, 
3. have content after that indent, and
4. have both a name and a type, separated by a colon

tests/data/format/numpydoc/numpydoc_section_ordering.py.out
DWesl: A variant on `textwrap.dedent` that ignored the indentation of the first line for determining common whitespace, which might work, then a similar variant of `textwrap.indent` at the end.

I would need to test how both handle
```python
    """\
    A summary line, indented the same as the others for all tools.

    Section Header
    --------------
    ...
    """
```

Another option might be to replace the `f"{quotes:s}{body:s}{quotes:s}"` reconstructing the new docstring with `f"{quotes:s}\\\n{body:s}{quotes:s}"` if the first character of the body is whitespace, but that still leaves the question of what to do with that "Returns" at the left margin.

pydocstringformatter/_formatting/base.py
DanielNoord: Yeah, let's only do the other branch! I didn't see it wasn't used anymore.

pydocstringformatter/_formatting/formatters_numpydoc.py
DanielNoord: I think a concise comment explaining this would be helpful for future us 😄

tests/data/format/numpydoc/numpydoc_section_ordering.py.out
DanielNoord: Can't we just check which line we are inserting on and if it's `0` we remove any indent and if higher we add an indent if it is missing?

DWesl: That's another option.  

The [numpydoc style guide](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) doesn't specify whether it prefers first line of docstring on same line as quotes or different line, but has [examples of both](https://numpydoc.readthedocs.io/en/latest/format.html#short-summary).  The [numpydoc validation tool](https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks) suggests the 
```python
    """\
    Summary.

    Extended summary...
    """
```
form is preferred to the
```python
    """Summary.

    Extended summary...
    """
```
form, which doesn't seem to agree with the style guide.  It also insists on extended summary and "See Also" sections, which I don't see always being necessary.

Supporting the first form is relatively straightforward; looping through the first lines of each section and doing `line.lstrip()` on the first section or `textwrap.indent(line)` for subsequent sections is also straightforward.

tests/data/format/numpydoc/numpydoc_header_line.py.out
DWesl: It matches the length of the word "Returns", and is shorter in the file by three characters, matching the three double quotes starting that line.

It might be clearer as
```python
def sincos(theta):
    """\
    Returns
    -------
    ...
    """
```
(same number of hyphens)

pydocstringformatter/_formatting/base.py
```diff
@@ -202,3 +206,96 @@ def treat_summary(
...
+        # Split sections
+        lines = tokeninfo.string[quotes_length:-quotes_length].split("\n")
+        # handle the spaces before the closing quotes
+        last_line = lines[-1]
+        if lines[-1].isspace():
+            lines[-1] = ""
+
+        # Everything before the first section header is in a single
+        # summary/deprecation warning/extended summary section.  This
+        # ends up being called "Summary".
```
DanielNoord, line 246:
```suggestion
```

Now that I look at it again, this no longer makes sense here.

pydocstringformatter/_formatting/base.py
```diff
@@ -202,3 +206,96 @@ def treat_summary(
...
+        new_sections = self.treat_sections(sections)
+
+        # Check that indent on first line of section didn't get weird
+        first_section = True
+        for section in new_sections.values():
+            if first_section:
+                section[0] = section[0].lstrip()
+            elif not section[0][0].isspace():
+                section[0] = f"{' ' * indent_length:s}{section[0]:s}"
+            first_section = False
```
DanielNoord, line 287:
```suggestion
                section[0] = section[0].lstrip()
                first_section = False
            elif not section[0][0].isspace():
                section[0] = f"{' ' * indent_length:s}{section[0]:s}"            
```

Saves some reassignments 😄

tests/data/format/numpydoc/numpydoc_header_line.py.out
DanielNoord: I mean, there isn't really a style guide for this I think, but imo it makes sense to add more `-` here so that the second line covers both the quotes and `Returns`.

DanielNoord (changes requested):

DanielNoord (approved):

DanielNoord:
@DWesl Thanks for all the work you put into this PR. I'm going to make a release immediately after this has been merged 😄

PR #138, opened 2022-07-22
Add the ``--style`` option

DanielNoord:
Currently only supports ``--style pep257``

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -9,6 +9,7 @@ Current usage of ``pydocstringformatter``:
                                 [--exit-code] [--max-summary-lines int]
                                 [--summary-quotes-same-line]
                                 [--max-line-length int]
+                                [--style {pep257} [{pep257} ...]]
                                 [--strip-whitespaces  --no-strip-whitespaces]
                                 [--split-summary-body  --no-split-summary-body]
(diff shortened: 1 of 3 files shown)


docs/usage.rst
```diff
@@ -44,6 +45,9 @@ Current usage of ``pydocstringformatter``:
                             is enforced for single line docstrings.
       --max-line-length int
                             Maximum line length of docstrings.
+      --style {pep257} [{pep257} ...]
+                            Docstring styles that are used in the project. Can be
+                            more than one.
```
Pierre-Sassoulas, line 50: It means we'll have to detect which style is which given a docstring. We have something who does that in pylint, we might extract it to make something generic...

DanielNoord: Yeah that's the next step.

Although I don't really like the difficult regexes pylint uses. I'd rather build them from scratch here and see if we can keep them a little bit simpler.

We need less information that pylint needs to extract from the regexes anyway as we don't really care about the content of the docstrings...

Pierre-Sassoulas: > I'd rather build them from scratch here and see if we can keep them a little bit simpler.

Then it's an opportunity to remove the complicated regexes from pylint and instead of extracting we replace them 😄

Pierre-Sassoulas (approved):
👌

PR #145, opened 2022-08-10
Run formatters based on the selected style

DanielNoord:


DWesl:
With this framing, you could probably set the default for `--style` to `["default"]` and then append `"pep257"` if `len(self.namespace.styles) <= 1`

DanielNoord:
Yeah I thought about that. However, if users did `--style=pep257` `default` would be missing so they would need to do `--style=pep257,default`, which doesn't seem right.
Adding `default` to `style` would also work, but this if check accomplishes the same and might be clearer? I don't know honestly 😄

DWesl:
Depends on how the parser treats default values with `action="append"`.  I'd have to look into that to be sure.

DanielNoord:
Yeah it just appends to the default. So you could end up with `["default", "pep257", "default"]`. That's not necessarily any issue I guess.

DWesl:
> Yeah it just appends to the default. So you could end up with `["default", "pep257", "default"]`. That's not necessarily any issue I guess.

Since we're treating it as a set, not really.  If that bothers you, we could change the action to change the value to a `set` or `frozenset` when it's done parsing the command line.

Issue #155, opened 2022-09-07
Incorrect reading of pyproject.toml


jspaezp:
Hello there!

I found a bug where the 'style' argument is ignored from the configuration file

This is my the section of my toml

```toml
[tool.pydocstringformatter]
style = "randomdoc"
exclude = [".tox/**"]
max-line-length = 88
```

After a little lurking I think it is caused by this ....

The style arg is defined as an extend.

https://github.com/DanielNoord/pydocstringformatter/blob/7a973217b2f99c61557ce7d6ad5a6a8b042d96d4/pydocstringformatter/_configuration/arguments_manager.py#L121-L128

And extend style args are not checked in the parsing of the toml file. Thus replacing whatever is in the field with
an empty list (which gets ignored later)

https://github.com/DanielNoord/pydocstringformatter/blob/7a973217b2f99c61557ce7d6ad5a6a8b042d96d4/pydocstringformatter/_configuration/toml_parsing.py#L44-L52

Possible fix:

changing the `pydocstringformatter/_configuration/toml_parsing.py#L44-L52` to ...

```python
 if isinstance(action, argparse._StoreTrueAction): 
     if value is True: 
         return [action.option_strings[0]] 
     return [] 
 elif isinstance(action, argparse._StoreAction): 
     if isinstance(value, int): 
         value = str(value) 
     return [action.option_strings[0], value] 
 elif isinstance(action, argparse._ExtendAction): 
     if isinstance(value, str): 
         value = str(value) 
     return [action.option_strings[0], value] 
 else:
    raise NotImplementedError
# return []  # pragma: no cover
```

so it **1** Checks for the extend type args and **2** makes maintaining the package easier because it throws an error instead of silently ignored wrongly parsed files.

btw ... debugging was not super easy to be honest, is there a way to have the cli return the configuration that will be used? (explicit version of all the configuration arguments) ... like `--verbose` argument.

Let me know if you would like me to make PR for any of these

Best!
Sebastian

Pierre-Sassoulas:
Hello, thank you for the debugging and analyzing the issue ! PR are welcome, let's do one for each issue (easier debugging and the configuration bug) .

PR #156, opened 2022-09-07
fixed toml parser ignoring style

jspaezp:
Closes https://github.com/DanielNoord/pydocstringformatter/issues/155

### DIFF
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -41,15 +41,20 @@ def parse_toml_option(
         except KeyError:
             raise UnrecognizedOption(f"Don't recognize option {opt}") from exc
 
-    if isinstance(action, argparse._StoreTrueAction):
-        if value is True:
-            return [action.option_strings[0]]
-        return []
-    if isinstance(action, argparse._StoreAction):
-        if isinstance(value, int):
-            value = str(value)
-        return [action.option_strings[0], value]
-    return []  # pragma: no cover
+    if isinstance(action, argparse._StoreTrueAction): 
+        if value is True: 
+            return [action.option_strings[0]] 
+        return [] 
+    elif isinstance(action, argparse._StoreAction): 
+        if isinstance(value, int): 
+            value = str(value) 
+        return [action.option_strings[0], value] 
+    elif isinstance(action, argparse._ExtendAction): 
+        if isinstance(value, str): 
+            value = str(value) 
+        return [action.option_strings[0], value] 
+    else:
+       raise NotImplementedError
 
 
 def parse_toml_file(
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -52,8 +52,6 @@ def parse_toml_option(
         return [action.option_strings[0], value]
 
     if isinstance(action, argparse._ExtendAction):  # type: ignore[attr-defined]
-        if isinstance(value, str):
-            value = str(value)
         return [action.option_strings[0], value]
 
     raise NotImplementedError  # pragma: no cover
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -3,7 +3,11 @@ from __future__ import annotations
 import argparse
 import os
 import sys
-from argparse import _ExtendAction, _StoreAction, _StoreTrueAction  #  type: ignore[attr-defined]
+from argparse import (  # type: ignore[attr-defined]
+    _ExtendAction,
+    _StoreAction,
+    _StoreTrueAction,
+)
 from typing import Any
 
 from pydocstringformatter._utils.exceptions import TomlParsingError, UnrecognizedOption
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -57,7 +57,7 @@ def parse_toml_option(
             value = str(value)
         return [action.option_strings[0], value]
 
-    raise NotImplementedError # pragma: no cover
+    raise NotImplementedError  # pragma: no cover
 
 
 def parse_toml_file(
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -3,11 +3,6 @@ from __future__ import annotations
 import argparse
 import os
 import sys
-from argparse import (  # type: ignore[attr-defined]
-    _ExtendAction,
-    _StoreAction,
-    _StoreTrueAction,
-)
 from typing import Any
 
 from pydocstringformatter._utils.exceptions import TomlParsingError, UnrecognizedOption
@@ -46,17 +41,17 @@ def parse_toml_option(
         except KeyError:
             raise UnrecognizedOption(f"Don't recognize option {opt}") from exc
 
-    if isinstance(action, _StoreTrueAction):
+    if isinstance(action, argparse._StoreTrueAction):
         if value is True:
             return [action.option_strings[0]]
         return []
 
-    if isinstance(action, _StoreAction):
+    if isinstance(action, argparse._StoreAction):
         if isinstance(value, int):
             value = str(value)
         return [action.option_strings[0], value]
 
-    if isinstance(action, _ExtendAction):
+    if isinstance(action, argparse._ExtendAction):  # type: ignore[attr-defined]
         if isinstance(value, str):
             value = str(value)
         return [action.option_strings[0], value]
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import argparse
 import os
 import sys
-from argparse import _ExtendAction, _StoreAction, _StoreTrueAction
+from argparse import _ExtendAction, _StoreAction, _StoreTrueAction  #  type: ignore[attr-defined]
 from typing import Any
 
 from pydocstringformatter._utils.exceptions import TomlParsingError, UnrecognizedOption
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import os
 import sys
+from argparse import _ExtendAction, _StoreAction, _StoreTrueAction
 from typing import Any
 
 from pydocstringformatter._utils.exceptions import TomlParsingError, UnrecognizedOption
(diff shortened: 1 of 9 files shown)


Pierre-Sassoulas (changes requested):
Thank you for opening a pull request. Could you add covering tests, please ?

jspaezp:
@Pierre-Sassoulas Sure thing! Done! Added testing that checks if the style is applied correctly depending on the toml file.

Regarding the the added lines marked as not being covered (not implemented error) by definition cannot be covered.
It will be raised if the implementation of the arguments changes in a way that is not supported by the parser (without it, it will fail silently instead of raising an error)

DanielNoord:
> Regarding the the added lines marked as not being covered (not implemented error) by definition cannot be covered. It will be raised if the implementation of the arguments changes in a way that is not supported by the parser (without it, it will fail silently instead of raising an error)

You can add a `# pragma: no cover` comment to this line to force the coverage to pass there. 

Could you also take a look at the failing `pre-commit` run?

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -45,11 +45,16 @@ def parse_toml_option(
         if value is True:
             return [action.option_strings[0]]
         return []
-    if isinstance(action, argparse._StoreAction):
+    elif isinstance(action, argparse._StoreAction):
```
Pierre-Sassoulas, line 48:
```suggestion
    if isinstance(action, argparse._StoreAction):
```
There's a return in each branch in the preceding ``if``. (that's why pre-commit fail.)

jspaezp: I see, I was not focusing in that error, noticed the one thrown by mypy

jspaezp:
> > Regarding the the added lines marked as not being covered (not implemented error) by definition cannot be covered. It will be raised if the implementation of the arguments changes in a way that is not supported by the parser (without it, it will fail silently instead of raising an error)
> 
> You can add a `# pragma: no cover` comment to this line to force the coverage to pass there.
> 
> Could you also take a look at the failing `pre-commit` run?

I am not sure why the mypy pre-commit is failing, it fails with error:

```
pydocstringformatter/_configuration/toml_parsing.py:52: error: Module has no attribute "_ExtendAction"; maybe "_AppendAction"?  [attr-defined]
Found 1 error in 1 file (checked 37 source files)
```

which is odd ... since it has been added since python 3.8 (and the fact that the code runs, if the object did not exist it would result in a runtime error):

https://github.com/python/cpython/commit/aa32a7e1116f7aaaef9fec453db910e90ab7b101

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -3,6 +3,7 @@
 import argparse
 import os
 import sys
+from argparse import _ExtendAction, _StoreAction, _StoreTrueAction
```
Pierre-Sassoulas, line 6:
```suggestion
from argparse import _ExtendAction, _StoreAction, _StoreTrueAction  #  type: ignore[attr-defined]
```

Pierre-Sassoulas (changes requested):
LGTM, we need to silence the mypy error if it's passing all the test imo.

Pierre-Sassoulas (approved):

DanielNoord:
> LGTM, we need to silence the mypy error if it's passing all the test imo.

Hmm, I don't really agree. It seems as if the `pre-commit` environment uses `python3`. That could be an older version.

We should at least try setting the `pre-commit.ci` environment to use `3.10` and see if that helps.

Pierre-Sassoulas:
> We should at least try setting the pre-commit.ci environment to use 3.10 and see if that helps.

I think that's the default right now ? Otherwise the option is ``language_version: python3.10``

jspaezp:
I have tried several configurations to set the python version and none seems to fix the error.
Even locally installing mypy and passing the arguments ...

```
$   python -m mypy --config-file=pyproject.toml
pydocstringformatter/_configuration/toml_parsing.py:6: error: Module "argparse" has no attribute "_ExtendAction"; maybe "_AppendAction"?  [attr-defined]
pydocstringformatter/_testutils/primer/packages.py:57: error: Module has no attribute "Repo"  [attr-defined]
Found 2 errors in 2 files (checked 34 source files)

$   python -m mypy --config-file=pyproject.toml --python-version=3.10
pydocstringformatter/_configuration/toml_parsing.py:6: error: Module "argparse" has no attribute "_ExtendAction"; maybe "_AppendAction"?  [attr-defined]
pydocstringformatter/_testutils/primer/packages.py:57: error: Module has no attribute "Repo"  [attr-defined]
Found 2 errors in 2 files (checked 34 source files)

$   python -m mypy --version                                      
mypy 0.971 (compiled: yes)

$  python --version
Python 3.10.4

$   python -c "import argparse ; print(argparse._ExtendAction)"
<class 'argparse._ExtendAction'>

$   python -c "import git ; print(git.Repo)"
<class 'git.repo.base.Repo'>
```

DanielNoord:
Oh wow, it seems like `typeshed` hasn't been updated for this.

See https://github.com/python/typeshed/blob/master/stdlib/argparse.pyi.

If you want you can open an issue or PR for this against `typeshed`. I think they would appreciated that.

I'll try and get a review of this PR tonight. Thanks for the testing you already did @jspaezp!

jspaezp:
Wow! I did not know about typeshed!

Just for reference in the future:
PR adding `_ExtendAction` to typeshed:
https://github.com/python/typeshed/pull/8701

DanielNoord:
> Wow! I did not know about typeshed!
> 
> Just for reference in the future: PR adding `_ExtendAction` to typeshed: [python/typeshed#8701](https://github.com/python/typeshed/pull/8701)
> 
> It got approved! so in time mypy will work correctly

Congratulations on your first contribution! 🎉

Pierre-Sassoulas:
> It got approved! so in time mypy will work correctly

This is also going to impact pylint 😄

https://github.com/python/typeshed/pull/8701#issuecomment-1239861775

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -45,11 +45,18 @@ def parse_toml_option(
...
             return [action.option_strings[0]]
         return []
+
     if isinstance(action, argparse._StoreAction):
         if isinstance(value, int):
             value = str(value)
         return [action.option_strings[0], value]
-    return []  # pragma: no cover
+
+    if isinstance(action, argparse._ExtendAction):  # type: ignore[attr-defined]
```
DanielNoord, line 54: I added the `argparse.` notation as we also have some actions that don't come from `argparse`. This makes this a bit clearer.

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -45,11 +45,18 @@ def parse_toml_option(
...
+        if isinstance(value, str):
```
DanielNoord, line 55: Shouldn't this be `if not`? Why are we casting a `str` to a `str`?

tests/test_config.py
```diff
@@ -62,6 +62,42 @@ def test_valid_toml_two(
     assert not output.err
 
 
+def test_toml_style(
+    capsys: pytest.CaptureFixture[str], monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """Test a correct toml with write = False."""
```
DanielNoord, line 68: This needs to be updated I think

tests/test_config.py
```diff
@@ -62,6 +62,42 @@ def test_valid_toml_two(
...
+    monkeypatch.chdir(CONFIG_DATA / "valid_toml_two")
+    pydocstringformatter.run_docstring_formatter(["test_package"])
```
DanielNoord, line 70: See comments below, but I think it's better if you use `_Run` here as you can then do:

`run = _Run(["test_package"])`
`assert run.config == ...`

The tests should also be moved to the `TestStyleOption` which gives other examples of how to do this 😄

tests/test_config.py
```diff
@@ -62,6 +62,42 @@ def test_valid_toml_two(
...
+'''
+    )
+
+    with open("pyproject.toml", "a", encoding="utf8") as file:
+        file.write("style = 'numpydoc'")
+
+    pydocstringformatter.run_docstring_formatter(["test_package"])
+    numpydoc_output = capsys.readouterr()
+
+    assert numpydoc_output.out.endswith(
```
DanielNoord, line 89: Thanks for spotting this issue!

For this docstring `numpydoc` and `pep257` shouldn't have different output. I have pushed https://github.com/DanielNoord/pydocstringformatter/pull/158 to address this.
Could you merge `main` into your branch and update this test? I think taking inspiration from `TestStyleOption` would be best. Those tests show how to check that the `config` attribute is set correctly.

DanielNoord (changes requested):
Thank you very much @jspaezp for all the work here.

Sadly this PR unearthed another issue within the tool that (after being addressed) will break the current test. I have given some pointers to making a new test. If you need any help please let me know!

pydocstringformatter/_configuration/toml_parsing.py
jspaezp: actually ... it might not be needed at all, will fix that

tests/data/config/valid_toml_numpydoc/test_package/numpydoc_style.py
```diff
@@ -0,0 +1 @@
+../../../../data/format/numpydoc/numpydoc_regression.py
```
DanielNoord, line 1: Could you just copy the contents of the file here instead of using a symlink? If we ever change the directory structure I can see myself wondering for 5 minutes why that would affect unrelated tests like this 😄

tests/test_config.py
```diff
@@ -255,3 +255,13 @@ def test_style_invalid_choice(self, capsys: pytest.CaptureFixture[str]) -> None:
         output = capsys.readouterr()
         assert not output.out
         assert "--style: invalid choice: 'invalid'" in output.err
+
+    def test_style_in_toml(self, monkeypatch: pytest.MonkeyPatch) -> None:
```
DanielNoord, line 259: Wonderful!

Should we also test for multiple styles? We support that on the command line so we should probably support that in `toml` as well?

DanielNoord (changes requested):

tests/data/config/valid_toml_numpydoc/test_package/numpydoc_style.py
jspaezp: Will do!

tests/test_config.py
jspaezp: done

tests/test_config.py
```diff
@@ -255,3 +255,17 @@ def test_style_invalid_choice(self, capsys: pytest.CaptureFixture[str]) -> None:
...
+        run = _Run(["test_package"])
+        assert ["numpydoc"] == run.config.style
+
+        monkeypatch.chdir(CONFIG_DATA / "valid_toml_pep257")
+        run = _Run(["test_package"])
+        assert ["pep257"] == run.config.style
+
+        monkeypatch.chdir(CONFIG_DATA / "valid_toml_numpydoc_pep257")
+        run = _Run(["test_package"])
+        assert all(x in run.config.style for x in ("pep257", "numpydoc"))
```
DanielNoord, line 271:
```suggestion
        assert run.config.style == ["pep257", "numpydoc"]
```

tests/test_config.py
```diff
@@ -255,3 +255,17 @@ def test_style_invalid_choice(self, capsys: pytest.CaptureFixture[str]) -> None:
...
+        run = _Run(["test_package"])
+        assert ["numpydoc"] == run.config.style
+
+        monkeypatch.chdir(CONFIG_DATA / "valid_toml_pep257")
+        run = _Run(["test_package"])
+        assert ["pep257"] == run.config.style
+
+        monkeypatch.chdir(CONFIG_DATA / "valid_toml_numpydoc_pep257")
+        run = _Run(["test_package"])
+        assert run.config.style == ["pep257", "numpydoc"]
```
DanielNoord, line 271:
```suggestion
        assert run.config.style == ["numpydoc", "pep257"]
```

DanielNoord (approved):

DanielNoord:
Thanks @jspaezp 😄

Issue #159, opened 2022-09-09
Bug, Index error when the last summary line is empty


jspaezp:
Noticed that the software breaks when the last line of the summary line is empty

Breaking args:

        # Breaks >>> Call Args:  ['tests/data/format/quotes_type/module_docstring.py', '--style', 'numpydoc', '--strip-whitespaces']
        # Works >>> Call Args:  ['tests/data/format/quotes_type/module_docstring.py', '--style', 'numpydoc', '--no-strip-whitespaces']

Solution:

Modifying:

https://github.com/DanielNoord/pydocstringformatter/blob/c749fc62699443a784e3211ed9aa0d591134db6a/pydocstringformatter/_formatting/_utils.py#L1-L9

To:

```python
 import functools 
  
  
 @functools.lru_cache(maxsize=None) 
 def is_rst_title(summary: str) -> bool: 
     """Check if the second line of a summary is one recurring character.""" 
     # If second line is one recurring character we're dealing with a rst title 
     if len(last_line) == 0:
        return False
     last_line = summary.splitlines()[-1].lstrip() 
     return last_line.count(last_line[0]) == len(last_line) 

```

DanielNoord:
I don't think your solution would work as `last_line` won't be defined at the `len` check, but please open a PR to fix the issue. This is indeed a bug.

jspaezp:
my bad, it was copy/paste from my local fix

PR #161, opened 2022-09-10
Bugfix/emptyline no whitespace

jspaezp:
Fixes: https://github.com/DanielNoord/pydocstringformatter/issues/159

### DIFF
--- a/pydocstringformatter/_formatting/_utils.py
+++ b/pydocstringformatter/_formatting/_utils.py
@@ -6,6 +6,6 @@ def is_rst_title(summary: str) -> bool:
     """Check if the second line of a summary is one recurring character."""
     # If second line is one recurring character we're dealing with a rst title
     last_line = summary.splitlines()[-1].lstrip()
-    if len(last_line) == 0:
+    if len(last_line) < 1:
         return False
     return last_line.count(last_line[0]) == len(last_line)
--- a/pydocstringformatter/_formatting/_utils.py
+++ b/pydocstringformatter/_formatting/_utils.py
@@ -6,6 +6,6 @@ def is_rst_title(summary: str) -> bool:
     """Check if the second line of a summary is one recurring character."""
     # If second line is one recurring character we're dealing with a rst title
     last_line = summary.splitlines()[-1].lstrip()
-    if len(last_line) < 1:
+    if not last_line:
         return False
     return last_line.count(last_line[0]) == len(last_line)
--- a/pydocstringformatter/_formatting/_utils.py
+++ b/pydocstringformatter/_formatting/_utils.py
@@ -6,4 +6,6 @@ def is_rst_title(summary: str) -> bool:
     """Check if the second line of a summary is one recurring character."""
     # If second line is one recurring character we're dealing with a rst title
     last_line = summary.splitlines()[-1].lstrip()
+    if len(last_line) == 0:
+        return False
     return last_line.count(last_line[0]) == len(last_line)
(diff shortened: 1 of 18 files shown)


tests/data/format/no_whitespace_stripper/module_docstring.py.out
```diff
@@ -0,0 +1,18 @@
+"""  A multi-line  .
```
Pierre-Sassoulas, line 1: I wonder if we should rstrip the end of the line before adding the final dot. Thought @DanielNoord ?

Pierre-Sassoulas (commented):
Looks very nice, a straightforward fix and a good accompanying test, thank you.

tests/data/format/no_whitespace_stripper/module_docstring.py.out
jspaezp: I was thinking the same thing while writing the test, but I figured at the end of the day the argument is called `--no-strip-whitespace` .... so I figured it was consistent in its behavior.

Maybe down the road we could change the argument such that there is the possibility of only removing trailing whitespace, instead of always removing both trailing and starting.

LMK what you think

pydocstringformatter/_formatting/_utils.py
```diff
@@ -6,4 +6,6 @@ def is_rst_title(summary: str) -> bool:
     """Check if the second line of a summary is one recurring character."""
     # If second line is one recurring character we're dealing with a rst title
     last_line = summary.splitlines()[-1].lstrip()
+    if len(last_line) < 1:
```
DanielNoord, line 9:
```suggestion
    if not last_line:
```

This should also work and save the call to `len`!

tests/data/format/no_whitespace_stripper/module_docstring.py.out
DanielNoord: Yeah agreed.

I think if you're letting a tool put periods in your docstrings automatically you'll likely also let that tool remove whitespaces. So they probably won't be without the other very often. And in those cases where people do I think it is good to offer the flexibility.

tests/data/format/no_whitespace_stripper/module_docstring.py.out
```diff
@@ -0,0 +1,18 @@
...
+docstring  
+   
+"""
+
+"""  My docstring.
+
+ 
+    
+   My indented section  
+
```
DanielNoord, line 17: This seems to be a very weird edge case. Shouldn't L17 have as many whitespaces as L12 has in the original file?

DanielNoord, line 17: This seems to be a very weird edge case. Shouldn't L17 have as many whitespaces as L12 has in the original file?

jspaezp: I think It does ...

(screenshot from my vscode just to render the whitespace ...)
image

jspaezp: I believe it does (at least when looking at it in vscode to render the whitespace)

pydocstringformatter/_formatting/_utils.py
```diff
@@ -5,5 +5,7 @@
 def is_rst_title(summary: str) -> bool:
     """Check if the second line of a summary is one recurring character."""
     # If second line is one recurring character we're dealing with a rst title
-    last_line = summary.splitlines()[-1].lstrip()
+
```
DanielNoord, line 8:
```suggestion
```

tests/data/format/no_whitespace_stripper/module_docstring.py.out
DanielNoord: Ah yeah, it does indeed. I should have checked the raw file myself.

DanielNoord (approved):

DanielNoord (approved):

Issue #160, opened 2022-09-09
Bug, Unhandled case when a section is empty


jspaezp:
# Issue

    # >>> Call Args:  ['tests/data/format/numpydoc/numpydoc_parameter_spacing.py', '--style', 'numpydoc', '--style', 'pep257']

This call fails with an index error because `section[0]` has length 0, therefore `section[0][0]` returns an index error.

# Fix

Modifying:

https://github.com/DanielNoord/pydocstringformatter/blob/c749fc62699443a784e3211ed9aa0d591134db6a/pydocstringformatter/_formatting/base.py#L275-L282

To:

```python
 # Check that indent on first line of section didn't get weird 
 first_section = True 
 for section in new_sections.values(): 
     if first_section: 
         section[0] = section[0].lstrip() 
         first_section = False 
     elif len(section[0]) > 0 and not section[0][0].isspace(): # <<<<<< this is the fix
         section[0] = f"{' ' * indent_length:s}{section[0]:s}" 
```

## Related:

Breaking call:

    # >>> Call Args:  ['tests/data/format/summary_splitter/regression_151.py', '--style', 'numpydoc', '--linewrap-full-docstring']

Fix:

https://github.com/DanielNoord/pydocstringformatter/blob/c749fc62699443a784e3211ed9aa0d591134db6a/pydocstringformatter/_formatting/formatters_default.py#L91-L101

```
        summary_lines = summary.splitlines()

        # Here is the fix, checks if the summary line is empty before going though
        if len(summary_lines) == 0:
            summary_lines = [""]

        new_summary = "\n".join(
            textwrap.wrap(
                summary_lines[0],
                width=line_length,
                initial_indent=" " * (indent_length + quotes_length),
                subsequent_indent=" " * indent_length,
                replace_whitespace=True,
            )
        )[indent_length + quotes_length :]


```

DanielNoord:
Sounds good! Could you open a PR with the proposed fix and a test?

DanielNoord:
Reopening as the test in https://github.com/DanielNoord/pydocstringformatter/pull/162 shows this can still reproduce in some edge cases.

I can't immediately think of a fix for https://github.com/DanielNoord/pydocstringformatter/pull/162 as it also needs to raise some sort of user warning like we saw in https://github.com/DanielNoord/pydocstringformatter/pull/278

PR #162, opened 2022-09-10
fix issue where passing two styles would fail with index error

jspaezp:
Fixes the first section of:

Closes https://github.com/DanielNoord/pydocstringformatter/issues/160

### DIFF
--- a/pydocstringformatter/_formatting/base.py
+++ b/pydocstringformatter/_formatting/base.py
@@ -278,7 +278,7 @@ class NumpydocSectionFormatter(StringAndQuotesFormatter, metaclass=abc.ABCMeta):
             if first_section:
                 section[0] = section[0].lstrip()
                 first_section = False
-            elif not section[0][0].isspace():
+            elif len(section[0]) > 0 and not section[0][0].isspace():
                 section[0] = f"{' ' * indent_length:s}{section[0]:s}"
 
         # Rejoin sections
--- a/pydocstringformatter/_formatting/base.py
+++ b/pydocstringformatter/_formatting/base.py
@@ -278,7 +278,7 @@ class NumpydocSectionFormatter(StringAndQuotesFormatter, metaclass=abc.ABCMeta):
             if first_section:
                 section[0] = section[0].lstrip()
                 first_section = False
-            elif len(section[0]) > 0 and not section[0][0].isspace():
+            elif section[0] and not section[0][0].isspace():
                 section[0] = f"{' ' * indent_length:s}{section[0]:s}"
 
         # Rejoin sections
--- a/tests/data/format/multi_style/numpydoc_parameter_spacing.py.out
+++ b/tests/data/format/multi_style/numpydoc_parameter_spacing.py.out
@@ -1,0 +1,74 @@
+"""Example module for numpydoc docstring style.
+
+References
+----------
+NumPy docstring style guide:
+https://numpydoc.readthedocs.io/en/latest/format.html#documenting-modules
+"""
+import math
+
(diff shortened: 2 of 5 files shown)


jspaezp:
Tests fail due to the auto-fix from pre-commit.ci 0c6795b437e01d62e2134eb8efd6099b59e54777
... which raises the question ... why is it generating those trailing whitespaces?
Should a "two pass" approach be done?

pydocstringformatter/_formatting/base.py
```diff
@@ -278,7 +278,7 @@ def treat_string(
             if first_section:
                 section[0] = section[0].lstrip()
                 first_section = False
-            elif not section[0][0].isspace():
+            elif len(section[0]) > 0 and not section[0][0].isspace():
```
DanielNoord, line 281:
```suggestion
            elif section[0] and not section[0][0].isspace():
```

DanielNoord (changes requested):

DanielNoord:
> * Should a "two pass" approach be done?

I have been thinking about this as well, but not sure if we should do that as we can create infinite loops.

Those trailing whitespaces seem to match the previous indentation. I think that is a bug and should only be a new line without the extra spacing.

jspaezp:
@DanielNoord What about this behavior? Where there is a section separator but no name. I think it makes clear that the parser was not able to assign what is under it to any section and not that it is part of the former.

> Those trailing whitespaces seem to match the previous indentation. I think that is a bug and should only be a new line without the extra spacing.

tests/data/format/multi_style/numpydoc_parameter_spacing.py.out
```diff
@@ -0,0 +1,74 @@
...
+
+References
+----------
+E (mathematical constant)
+https://en.wikipedia.org/wiki/E_(mathematical_constant)
+"""
+
+
+def sincos(theta):
+    """Returns.
```
DanielNoord, line 33: Oh this is becoming more difficult than I hoped. It seems both styles are somewhat incompatible with each other. This is the missing section name, which is now being used as a summary/title...

I'll have to think about this. Perhaps we should fix this in another PR, but I like to check if this is easily fixable.

jspaezp: Well ... if you believe that the issue is that the styles are mutually exclusive in some instances, enforcing that is not hard internally.

DanielNoord: I think the issue that they are currently mutually exclusive. Perhaps we can fix this by changing the order in which formatters are called. I hope to have time somewhere this week to look into this.

jspaezp:
@DanielNoord 

> > * Should a "two pass" approach be done?
> 
> I have been thinking about this as well, but not sure if we should do that as we can create infinite loops.
> 
> Those trailing whitespaces seem to match the previous indentation. I think that is a bug and should only be a new line without the extra spacing.

Thinking about this ... I started working a branch that does a two pass to check the stability of the tokens.
I will reference it in a new issue.

Pierre-Sassoulas:
Was this replaced by #169 @jspaezp ? (Should we close ?)

jspaezp:
Hello @Pierre-Sassoulas! sorry for not closing this issue before (just for the future, it was fixed here https://github.com/DanielNoord/pydocstringformatter/pull/163)

Pierre-Sassoulas:
No problem, thank you for answering :) !

DanielNoord:
I do think this test still fails on `main`? At least it did last time I checked after we merged #163.

PR #166, opened 2022-09-12
added boolean optional argument support

jspaezp:
Improved the toml parsing so it supports `BooleanOptionalAction` arguments in it.
(also added tests for it ...)

```toml
[tool.pydocstringformatter]
write = false
style = ["numpydoc"]
strip-whitespaces = true  # <<<<<- previously unsupported
split-summary-body = false # <<<<<- previously unsupported
no-numpydoc-section-hyphen-length = true # <<<<<- previously unsupported
```

.... where should I document this?

### DIFF
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -49,7 +49,8 @@ def parse_toml_option(
 
         if opt.startswith("no") and f"--{opt[3:]}" in option_strings:
             opposite_opt = opt[3:]
-            val, opp_val = ["false", "true"][value], ["true", "false"][value]
+            val = ["false", "true"][bool(value)]
+            opp_val = ["true", "false"][bool(value)]
             error_msg = (
                 "TOML file contains an unsupported option "
                 f"'{opt}: {val}', try using '{opposite_opt}: {opp_val}' instead"
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -49,7 +49,7 @@ def parse_toml_option(  # pylint: disable=too-many-branches
 
         if not isinstance(value, bool):
             error_msg = f"{{'{value}'}} {type(value)} is not a supported argument for"
-            error_msg += f"'{opt}', please use either {{true}} or {{false}}."
+            error_msg += f" '{opt}', please use either {{true}} or {{false}}."
             raise ValueError(error_msg)
 
         if opt.startswith("no") and f"--{opt[3:]}" in option_strings:
@@ -62,9 +62,7 @@ def parse_toml_option(  # pylint: disable=too-many-branches
             )
             raise TomlParsingError(error_msg)
 
-        index = option_strings.index(f"--{'no-' if not value else ''}{opt}")
-        option = action.option_strings[index]
-        return [option]
+        return [f"--{'no-' if not value else ''}{opt}"]
 
     if isinstance(action, argparse._StoreTrueAction):
         if value is True:
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -45,14 +45,12 @@ def parse_toml_option(  # pylint: disable=too-many-branches
             raise UnrecognizedOption(f"Don't recognize option {opt}") from exc
 
     if isinstance(action, BooleanOptionalAction):
-        option_strings = action.option_strings
-
         if not isinstance(value, bool):
             error_msg = f"{{'{value}'}} {type(value)} is not a supported argument for"
             error_msg += f" '{opt}', please use either {{true}} or {{false}}."
             raise ValueError(error_msg)
 
-        if opt.startswith("no") and f"--{opt[3:]}" in option_strings:
+        if opt.startswith("no") and f"--{opt[3:]}" in action.option_strings:
             opposite_opt = opt[3:]
             val = ["false", "true"][value]
             opp_val = ["true", "false"][value]
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -5,6 +5,9 @@ import os
 import sys
 from typing import Any
 
+from pydocstringformatter._configuration.boolean_option_action import (
+    BooleanOptionalAction,
+)
 from pydocstringformatter._utils.exceptions import TomlParsingError, UnrecognizedOption
 
 if sys.version_info <= (3, 11):
@@ -40,6 +43,11 @@ def parse_toml_option(
             action = parser._option_string_actions[f"-{opt}"]
         except KeyError:
             raise UnrecognizedOption(f"Don't recognize option {opt}") from exc
+
+    if isinstance(action, BooleanOptionalAction):
+        index = action.option_strings.index(f"--{opt}")
+        option = action.option_strings[index]
+        return [option]
 
     if isinstance(action, argparse._StoreTrueAction):
         if value is True:
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -45,7 +45,18 @@ def parse_toml_option(
             raise UnrecognizedOption(f"Don't recognize option {opt}") from exc
 
     if isinstance(action, BooleanOptionalAction):
-        index = action.option_strings.index(f"--{opt}")
+        option_strings = action.option_strings
+
+        if opt.startswith("no") and f"--{opt[3:]}" in option_strings:
+            opposite_opt = opt[3:]
+            val, opp_val = ["false", "true"][value], ["true", "false"][value]
+            error_msg = (
+                "TOML file contains an unsupported option "
+                f"'{opt}: {val}', try using '{opposite_opt}: {opp_val}' instead"
+            )
+            raise TomlParsingError(error_msg)
+
+        index = option_strings.index(f"--{'no-' if not value else ''}{opt}")
         option = action.option_strings[index]
         return [option]
 
--- a/pydocstringformatter/_configuration/toml_parsing.py
+++ b/pydocstringformatter/_configuration/toml_parsing.py
@@ -54,8 +54,8 @@ def parse_toml_option(  # pylint: disable=too-many-branches
 
         if opt.startswith("no") and f"--{opt[3:]}" in option_strings:
             opposite_opt = opt[3:]
-            val = ["false", "true"][bool(value)]
-            opp_val = ["true", "false"][bool(value)]
+            val = ["false", "true"][value]
+            opp_val = ["true", "false"][value]
             error_msg = (
                 "TOML file contains an unsupported option "
                 f"'{opt}: {val}', try using '{opposite_opt}: {opp_val}' instead"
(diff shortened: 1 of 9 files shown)


Pierre-Sassoulas (approved):
Looks good, thank you for adding this !

Pierre-Sassoulas:
This is the full doc on configuration: https://pydocstringformatter.readthedocs.io/en/latest/?badge=latest#configuration

Pierre-Sassoulas:
Maybe we could autogenerate the full list of possible options with default values ?

jspaezp:
I like that idea but it would not exemplify the use of 'no-xxxxx = true' style configs. ... Maybe also have comments showing the equivalent config on the negative form ??

Pierre-Sassoulas:
I think the no-x style is only used in CLI (or should) this is given by the `--help` command so we don't need to document it imo'

DanielNoord:
The no- options are needed to disable default checks. I think they should indeed be shown as a comment in the example config.

For pylint we use `tomlkit`. This is one of the few parsers for `toml` that saves comments and style.
A solution where we initialise all options and then write a file with `tomlkit` would be nice.
Perhaps it would be even better if we could create a `--create-config` command so that we can also offer it as a cli
tool?

Pierre-Sassoulas:
> The no- options are needed to disable default checks

Isn't it possible to do ``x=false`` or ``no-x=true`` ?

DanielNoord:
> > The no- options are needed to disable default checks
> 
> Isn't it possible to do `x=false` or `no-x=true` ?

It depends on how we handle that in this PR. We could just disallow them completely as well?

What do you think makes most sense @jspaezp?

jspaezp:
Hello there!

I thought about it for a while...
and I think the most consistent behavior would be to allow "--no" version. I really like the idea of the config being as close as possible to copy/pasting the command line arguments.

regarding: https://github.com/DanielNoord/pydocstringformatter/pull/166#issuecomment-1244981369
> Isn't it possible to do x=false or no-x=true ?

its easy but not SUUUUPER EASY, just having to consider it robust to arguments that in the future might start with --no- but are not of the same "optional boolean" type.

I will add the mods to the readme in the meantime

Pierre-Sassoulas:
I meant it's possible to do it as a user (so you should always be able to use simply 'x' which is simpler both for the user and for us). Imo we should not permit to use no-x in configuration, there's very little benefit and a lot of headaches to expect.

jspaezp:
Alright, I added the enforcement of the positive form of the arguments (+testing, +updated readme, +informative error message).

Pierre-Sassoulas (approved):
I like the clear error message ! LGTM, I'll let Daniel merge.

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -41,6 +44,22 @@ def parse_toml_option(
         except KeyError:
             raise UnrecognizedOption(f"Don't recognize option {opt}") from exc
 
+    if isinstance(action, BooleanOptionalAction):
+        option_strings = action.option_strings
+
+        if opt.startswith("no") and f"--{opt[3:]}" in option_strings:
+            opposite_opt = opt[3:]
+            val, opp_val = ["false", "true"][value], ["true", "false"][value]
```
DanielNoord, line 52: Can I break this by doing `no-numpydoc-section-hyphen-length = 1`? or `no-numpydoc-section-hyphen-length = yes`?

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -41,6 +44,22 @@ def parse_toml_option(
...
+            error_msg = (
+                "TOML file contains an unsupported option "
+                f"'{opt}: {val}', try using '{opposite_opt}: {opp_val}' instead"
+            )
+            raise TomlParsingError(error_msg)
+
+        index = option_strings.index(f"--{'no-' if not value else ''}{opt}")
```
DanielNoord, line 59: What is this looking up? The default value?

tests/test_config.py
```diff
@@ -269,3 +269,27 @@ def test_style_in_toml(self, monkeypatch: pytest.MonkeyPatch) -> None:
...
         run = _Run(["test_package"])
         assert run.config.style == ["numpydoc", "pep257"]
+
+    def test_boolopt_in_toml(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """Test that arguments of type BooleanOptionalAction work in toml files."""
+        monkeypatch.chdir(CONFIG_DATA / "valid_toml_boolopt")
+        run = _Run(["test_package"])
+        assert not run.config.summary_quotes_same_line
+
+        # dashes in the name dont allow the dot notation to get this value
```
DanielNoord, line 279:
```suggestion
        # dashes in the name dont allow the dot notation to get this value
```

Can't you use `_` like on L277?

tests/test_config.py
```diff
@@ -269,3 +269,27 @@ def test_style_in_toml(self, monkeypatch: pytest.MonkeyPatch) -> None:
...
+        assert not run.config.__dict__["numpydoc-section-hyphen-length"]
+        assert run.config.__dict__["strip-whitespaces"]
+
+    def test_non_valid_boolopt_in_toml(self, monkeypatch: pytest.MonkeyPatch) -> None:
+        """Test that '--no' versions arguments in toml do not work."""
```
DanielNoord, line 284:
```suggestion
        """Test that '--no' versions of BooleanOptionalAction in toml do not work."""
```

DanielNoord (changes requested):

pydocstringformatter/_configuration/toml_parsing.py
jspaezp: It enforces the positive version regardless of the default value (it also does not assume that the default is the first argument).

assuming the passed value is `eggs = false`
if the option strings are:

> ['--no-eggs', '-eggs'] --> index = 0
> ['--eggs', '--no-eggs'] --> index = 1

tests/test_config.py
jspaezp: They are registered differently ... cannot be done. I am not sure if there is any other less hacky way of getting that attribute.

This is what i get if i start a debugger in line 278
```python
(pdb) p run.config.__dict__.keys()
dict_keys([..., 'max_summary_lines', ... 'numpydoc-section-order', ...])
(pdb) p run.config.strip_whitespace
*** AttributeError: 'Namespace' object has no attribute 'strip_whitespace'
```

DanielNoord: Ah, you're finding so many issues with the current code 😅 

I think this is because of the way

pydocstringformatter/_configuration/toml_parsing.py
jspaezp:
```toml
no-numpydoc-section-hyphen-length = 1  # Pretty sure this would work fine
no-numpydoc-section-hyphen-length = "yes"  # Yes, I think it would break
```

could be sturdier by changing `value` to `bool(value)` in L52

DanielNoord: I'd prefer only accepting `true` or `false`. `bool("no")` would be `True` which is confusing imo.

tests/test_config.py
jspaezp: We can definitely note this and make an issue out of it (I don't feel like fixing it should be in the scope of this PR)

DanielNoord: I already did 😄 

See #167

pydocstringformatter/_configuration/toml_parsing.py
jspaezp: ok, i can see that, nontheless, that check would need to happen in L49, since by the time the program is in L52, it has already rejected the option and its only making an error message

what about ....
```python
# L49
if not isinstance(value, bool):
    raise ValueError(f"'{value}' is not a supported argument for '{opt}', please use either 'true' or 'false'")
# L50
```

DanielNoord: Perfect!

jspaezp: Done

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -41,6 +44,28 @@ def parse_toml_option(
         except KeyError:
             raise UnrecognizedOption(f"Don't recognize option {opt}") from exc
 
+    if isinstance(action, BooleanOptionalAction):
+        option_strings = action.option_strings
+
+        if not isinstance(value, bool):
+            error_msg = f"{{'{value}'}} {type(value)} is not a supported argument for"
```
DanielNoord, line 51:
```suggestion
            error_msg = f"{{'{value}'}} {type(value)} is not a supported argument for "
```

pydocstringformatter/_configuration/toml_parsing.py
```diff
@@ -41,6 +44,28 @@ def parse_toml_option(
...
+            error_msg += f"'{opt}', please use either {{true}} or {{false}}."
+            raise ValueError(error_msg)
+
+        if opt.startswith("no") and f"--{opt[3:]}" in option_strings:
+            opposite_opt = opt[3:]
+            val = ["false", "true"][bool(value)]
+            opp_val = ["true", "false"][bool(value)]
```
DanielNoord, line 58:
```suggestion
            val = ["false", "true"][value]
            opp_val = ["true", "false"][value]
```

This shouldn't be necessary anymore

pydocstringformatter/_configuration/toml_parsing.py
DanielNoord: But you're getting the index of a string in a list to then return the string at that index again?
Can't we just do `[f"--{'no-' if not value else ''}{opt}"]`?

DanielNoord (changes requested):

pydocstringformatter/_configuration/toml_parsing.py
jspaezp: I think the space is needed, we could move it to the next line.

```python
error_msg += f" '{opt}', please use either {{true}} or {{false}}."
#                        ^ here
```

pydocstringformatter/_configuration/toml_parsing.py
jspaezp: you are right

```python
index = option_strings.index(f"--{'no-' if not value else ''}{opt}")
option = action.option_strings[index]
return [option]
```

should become
```
option = f"--{'no-' if not value else ''}{opt}"
return [option]
```

pydocstringformatter/_configuration/toml_parsing.py
DanielNoord: I personally like spaces at the end of the line but either way is fine with me! Do what suits you 😄

pydocstringformatter/_configuration/toml_parsing.py
DanielNoord: You could even merge the lines and just return without the assignment in between. `return` vs. `option = ` has a shorter line length anyway

jspaezp: done in https://github.com/DanielNoord/pydocstringformatter/pull/166/commits/c51e3220b66f90f68c134cf0bd777ff82e214782

DanielNoord (approved):

DanielNoord:
Thanks once again @jspaezp 🎉 😄

PR #171, opened 2022-09-16
make black skip files in test data

jspaezp:


Pierre-Sassoulas (commented):
I suppose you're launching black as a standalone and not via pre-commit ? 

I think we need to be consistent with : https://github.com/DanielNoord/pydocstringformatter/blob/main/.pre-commit-config.yaml#L36, probably remove the exclude in pre-commit config ?

jspaezp:
I do both, just to prevent the whole 'commit >  fail pre commit > stage changes > commit again'.
Yeah I think we could move it out of pre-commit.

I am really liking the trend of having all configuration of the tooling in the `pyproject.toml`.
It would have the added benefit that if black is ran in the linting testing stage anywhere else, it would just read from the config (hypothetical case of moving the linting form pre_commit.ci to github actions or something like that)

Pierre-Sassoulas:
It look like pre-commit is not taking into account the pyproject.toml. (This is surprising but the pre-commit maintainer does have a grudge against  pyproject.toml now that I htink of it)

jspaezp:
that is indeed extremely interesting ... I guess it would require having it redundant (oddly enough it is only relevant to the CI, locally it does seem to read the configuration)

DanielNoord:
Could it be because we also use the `args` field? Does removing that make `black` take the file into account.

I believe `black` should be able to pick up a config file even through pre-commit.

jspaezp:
I do not believe that is it ... I was looking into black's source code and they only use the pyproject to update the default values of their parser, so I do not feel like passing both cli args and having a pyproject would break it....

BTW: I re-based the branch because the commit history was getting really dirty

DanielNoord:
@jspaezp I found the issue 😄 

See the issue I linked in the comment in the config file

DanielNoord (approved):

PR #172, opened 2022-09-16
added pyc and __pycache__ to gitignore

jspaezp:
added pyc and __pycache__ to gitignore

(I would personally add venv as well, since it is a very common practice to have the virtual environment in the same directory as the project)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 pydocstringformatter.egg-info
 dist/
 docs/_build
+
+**.pyc
+**/__pycache__
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ docs/_build
 
 **.pyc
 **/__pycache__
+venv/


Pierre-Sassoulas (approved):
We don't have the issue with ``venv`` because it's in our global gitignore I suppose. I'm also ok with adding it in gitignore.

jspaezp:
learning new stuff here ... adding stuff to my global gitignore now ...

Pierre-Sassoulas:
The global gitignore this for IDE specific file like ``.idea``, but at this point the name ``venv`` seems so ubiquitous in python dev that it's arguable it should be in all gitignores.

jspaezp:
ok .... knowing about global gitignores (thank you so much!!!) this PR seems a lot less relevant, it think it does not hurt to add them anyway

DanielNoord:
I don't like adding such ignores that are not project specific. What if the standard name of `venv` becomes `pythonenv` after a clash of names? Or what if the structure of virtual environment changes. Nobody will remember to update all gitignores to remove redundant lines. Furthermore, people can also use `pyenv` right now which removes the need for a `venv` directory. 
Case in point being `pylint` where we have some configuration files of which nobody knows if anybody is even using them anymore.

Any file or configuration that is needed to directly contribute to the project but is a result of local development environments should be ignored locally.

Issue #169, opened 2022-09-16
Solve/diagnose conflicting formatters


jspaezp:
Started discussion in PR #162 

https://github.com/DanielNoord/pydocstringformatter/pull/162#issuecomment-1248796789

Pitch: Passing over the formatters twice could help clear out problems where the work of one formatter is undone by another one.

The current implementation is mostly focused in finding that formatters conflict with each other (run again after a full pass)
and having it raise an error.

Current development branch ...
https://github.com/jspaezp/pydocstringformatter/tree/feature/twopass

Current broken tests ...

```
FAILED tests/test_config.py::TestStyleOption::test_style_in_toml - ValueError: Conflicting formatters: numpydoc-section-spacing, closing-quotes
FAILED tests/test_config.py::TestStyleOption::test_boolopt_in_toml - ValueError: Conflicting formatters: numpydoc-section-spacing, closing-quotes
FAILED tests/test_formatting.py::test_formatting[linewrap_summary-function_docstrings.py] - ValueError: Conflicting formatters: split-summary-body, final-period
FAILED tests/test_formatting.py::test_formatting[linewrap_summary-max_line_length_50.py] - ValueError: Conflicting formatters: split-summary-body, final-period
FAILED tests/test_formatting.py::test_formatting[numpydoc-numpydoc_regression.py] - ValueError: Conflicting formatters: numpydoc-section-spacing, closing-quotes
```

so .... is it of interest to add this changes? If anything it is a great debugging addition for development
should arguments be added to enable it?
should the error be more user-friendly?

If so and you have insights on those design aspects I will make this a PR

DanielNoord:
I haven't looked at your branch as I am on mobile, but I'm highly in favour of this change. I'll gladly help review a PR but sadly won't have time to work on this myself this week. 

Some pointers to consider:
- I think the maximum run count should be 2. 
- After the second run, a third run should occur. If we still create changes (note that these should ignore the --write flag but should only check if there are any potential change) we need to emit an error. `black` has a very elegant system where after there is an inconsistency aftwr 2/3 runs a pre-regenerated bug report is send to stdout. This can then be copied to an issue in GitHub which would allow us to resolve the issue.
- We should test the issue template. So we need to write two incompatible test formatters and write a test where they fail to find a solution.

That's all I can think of for now. Let me know if you have any questions! 😊

PR #173, opened 2022-09-16
Feature: two-pass formatting of the tokens

jspaezp:
Discussed here:
https://github.com/DanielNoord/pydocstringformatter/issues/169

Things to iron out:
- [ ] should arguments be added to enable it? (--num-passes ... --no-check ????) .... I vote no arguments ...
- [x] should the error be more user-friendly?
- [ ] Regarding the failing tests ... option A. Fix the formatters so they become compatible, option B. Fix the tests so the incompatible formatters do not get called together (also change the defauts of just adding the --style"numpydoc" to prevent automatic collisions ...). LMK what you think.
- [x] testing ... Right not by "playing with it" it works but a better unit testing is required.

https://github.com/DanielNoord/pydocstringformatter/issues/169#issuecomment-1249090803

> I haven't looked at your branch as I am on mobile, but I'm highly in favour of this change. I'll gladly help review a PR but sadly won't have time to work on this myself this week.
> 
> Some pointers to consider:
> 
>     * I think the maximum run count should be 2.
> 
>     * After the second run, a third run should occur. If we still create changes (note that these should ignore the --write flag but should only check if there are any potential change) we need to emit an error. `black` has a very elegant system where after there is an inconsistency aftwr 2/3 runs a pre-regenerated bug report is send to stdout. This can then be copied to an issue in GitHub which would allow us to resolve the issue.
> 
>     * We should test the issue template. So we need to write two incompatible test formatters and write a test where they fail to find a solution.
> 
> 
> That's all I can think of for now. Let me know if you have any questions! 😊

- [ ] Issue template to report the problems

Current error:

```
>           raise RuntimeError(msg)
E           RuntimeError: Conflicting formatters: numpydoc-section-spacing, closing-quotes
E           --- numpydoc-section-spacing vs closing-quotes /private/var/folders/5j/x2qvf77j1m7736lccs2xt6c40000gn/T/pytest-of-sebastianpaez/pytest-187/test_formatting_numpydoc_numpy0/test_file.py
E           +++ numpydoc-section-spacing vs closing-quotes /private/var/folders/5j/x2qvf77j1m7736lccs2xt6c40000gn/T/pytest-of-sebastianpaez/pytest-187/test_formatting_numpydoc_numpy0/test_file.py
E           @@ -1,2 +1 @@
E           -"""A docstring.
E           -    """
E           +"""A docstring."""
```

### DIFF
--- a/pydocstringformatter/_utils/issue_template.py
+++ b/pydocstringformatter/_utils/issue_template.py
@@ -1,0 +1,78 @@
+from __future__ import annotations
+
+import tokenize
+
+from pydocstringformatter._formatting import Formatter
+from pydocstringformatter._utils.file_diference import compare_formatters
+
+
+def create_gh_issue_template(
+    token: tokenize.TokenInfo, formatters: dict[str, Formatter], filename: str
+) -> str:
+    """Make a template for a GitHub issue.
+
+    Args:
+        token: The token that caused the issue.
+        formatters: The formatters that caused the issue.
+        filename: The filename of the file that caused the issue.
+    """
+    formatter_names = list(formatters)
+    msg = ""
+    if len(formatter_names) > 2:
+        msg = f"""
+Conflicting formatters: {", ".join(formatters)}
+"""
+        diff = f"Diff too intricate to compute for {len(formatter_names)} formatters."
+    else:
+        if len(formatter_names) == 2:
+            msg = f"""
+Conflicting formatters: {" and ".join(formatter_names)}
+These formatters conflict with each other for:
+
+```python
+{token.string}
+```
+"""
+            formatter_1 = formatters[formatter_names[0]]
+            formatter_2 = formatters[formatter_names[1]]
+        else:
+            msg = f"""
+Formatter: {formatter_names[0]}
+This formatter is not able to make stable changes for:
+
+```python
+{token.string}
+```
+"""
+            formatter_1 = formatters[formatter_names[0]]
+            formatter_2 = formatter_1
+
+        diff = compare_formatters(
+            token,
+            formatter_1,
+            formatter_2,
+            title_extra=str(filename),
+        )
+
+    out = f"""
+Unfortunately an error occurred while formatting a docstring.
+Please help us fix this bug by opening an issue at:
+https://github.com/DanielNoord/pydocstringformatter/issues/new
+
+{"-" * 80}
+
+You can use the following template when you open the issue:
+
+# Description:
+
+{msg}
+
+# Diff:
+
+```diff
+{diff}
+```
+
+"""
+
+    return out
--- a/pydocstringformatter/run.py
+++ b/pydocstringformatter/run.py
@@ -9,6 +9,7 @@ from pathlib import Path
 
 from pydocstringformatter import __version__, _formatting, _utils
 from pydocstringformatter._configuration.arguments_manager import ArgumentsManager
+from pydocstringformatter._utils.exceptions import UnstableResultError
 
 
 class _Run:
@@ -64,7 +65,7 @@ class _Run:
             # the same later on.
             newlines = file.newlines
 
-        formatted_tokens, is_changed = self.format_file_tokens(tokens)
+        formatted_tokens, is_changed = self.format_file_tokens(tokens, filename)
 
         if is_changed:
             try:
@@ -112,9 +113,22 @@ class _Run:
         return enabled
 
     def format_file_tokens(
-        self, tokens: list[tokenize.TokenInfo]
+        self, tokens: list[tokenize.TokenInfo], filename: Path
     ) -> tuple[list[tokenize.TokenInfo], bool]:
-        """Format a list of tokens."""
+        """Format a list of tokens.
+
+        tokens: List of tokens to format.
+        filename: Name of the file the tokens are from.
+
+        Returns:
+            A tuple containing [1] the formatted tokens in a list
+            and [2] a boolean indicating if the tokens were changed.
+
+        Raises:
+            UnstableResultError::
+                If the formatters are not able to get to a stable result.
+                It reports what formatters are still modifying the tokens.
+        """
         formatted_tokens: list[tokenize.TokenInfo] = []
         is_changed = False
 
@@ -122,20 +136,68 @@ class _Run:
             new_tokeninfo = tokeninfo
 
             if _utils.is_docstring(new_tokeninfo, tokens[index - 1]):
-                for _, formatter in self.enabled_formatters.items():
-                    new_tokeninfo = formatter.treat_token(new_tokeninfo)
-            formatted_tokens.append(new_tokeninfo)
+                new_tokeninfo, changers = self.apply_formatters(new_tokeninfo)
+                is_changed = is_changed or bool(changers)
 
-            if tokeninfo != new_tokeninfo:
-                is_changed = True
+                # Run formatters again (3rd time) to check if the result is stable
+                _, changers = self._apply_formatters_once(
+                    new_tokeninfo,
+                )
+
+                if changers:
+                    conflicting_formatters = {
+                        k: v
+                        for k, v in self.enabled_formatters.items()
+                        if k in changers
+                    }
+                    template = _utils.create_gh_issue_template(
+                        new_tokeninfo, conflicting_formatters, str(filename)
+                    )
+
+                    raise UnstableResultError(template)
+
+            formatted_tokens.append(new_tokeninfo)
 
         return formatted_tokens, is_changed
 
+    def apply_formatters(
+        self, token: tokenize.TokenInfo
+    ) -> tuple[tokenize.TokenInfo, set[str]]:
+        """Apply the formatters twice to a token.
+
+        Also tracks which formatters changed the token.
+
+        Returns:
+            A tuple containing:
+            [1] the formatted token and
+            [2] a set of formatters that changed the token.
+        """
+        token, changers = self._apply_formatters_once(token)
+        if changers:
+            token, changers2 = self._apply_formatters_once(token)
+            changers.update(changers2)
+        return token, changers
+
+    def _apply_formatters_once(
+        self, token: tokenize.TokenInfo
+    ) -> tuple[tokenize.TokenInfo, set[str]]:
+        """Applies formatters to a token and keeps track of what changes it.
+
+        token: Token to apply formatters to
+
+        Returns:
+            A tuple containing [1] the formatted token and [2] a set
+            of formatters that changed the token.
+        """
+        changers: set[str] = set()
+        for formatter_name, formatter in self.enabled_formatters.items():
+            if (new_token := formatter.treat_token(token)) != token:
+                changers.add(formatter_name)
+                token = new_token
+
+        return token, changers
+
     def format_files(self, filepaths: list[Path]) -> bool:
         """Format a list of files."""
-        is_changed = False
-
-        for file in filepaths:
-            is_changed = self.format_file(file) or is_changed
-
-        return is_changed
+        is_changed = [self.format_file(file) for file in filepaths]
+        return any(is_changed)
--- a/tests/data/format/linewrap_summary/function_docstrings.py.out
+++ b/tests/data/format/linewrap_summary/function_docstrings.py.out
@@ -1,5 +1,6 @@
 def func():
-    """A very long summary line that needs to be wrapped, A very long summary line that
+    """A very long summary line that needs to be wrapped, A very long summary line that.
+
     needs to be wrapped.
 
     A description that is not too long.
@@ -7,7 +8,8 @@ def func():
 
 
 def func():
-    """A very long multi-line summary line that needs to be wrapped, A very long multi-
+    """A very long multi-line summary line that needs to be wrapped, A very long multi-.
+
     line summary line that needs to be wrapped.
 
     A very long summary line that needs to be wrapped.
@@ -28,13 +30,15 @@ def func():
 # Since the ending quotes will be appended on the same line this
 # exceeds the max length.
 def func():
-    """A multi-line summary that can be on one line, Something that is just too
+    """A multi-line summary that can be on one line, Something that is just too.
+
     longgg.
     """
 
 
 def func():
-    """A multi-line summary that can be on one line, Something that is just too
+    """A multi-line summary that can be on one line, Something that is just too.
+
     long.
     """
 
@@ -46,6 +50,7 @@ def func():
 # Regression for bug found in pylint
 # We should re-add the quotes to line length if they will never be on the first line.
 class LinesChunk:
-    """The LinesChunk object computes and stores the hash of some consecutive stripped
+    """The LinesChunk object computes and stores the hash of some consecutive stripped.
+
     lines of a lineset.
     """
--- a/tests/test_conflicting_formatters.py
+++ b/tests/test_conflicting_formatters.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+from collections.abc import Generator
+from contextlib import contextmanager
+from pathlib import Path
+
+import pytest
+
+from pydocstringformatter import _formatting
+from pydocstringformatter import _testutils as test_utils
+from pydocstringformatter._formatting import Formatter
+from pydocstringformatter._utils import UnstableResultError
+from pydocstringformatter.run import _Run
+
+
+@contextmanager
+def patched_run(formatters: list[Formatter]) -> Generator[type[_Run], None, None]:
+    """Patches formatters so Run only uses the provided formatters."""
+    old_formatters = _formatting.FORMATTERS
+    _formatting.FORMATTERS = formatters
+    yield _Run
+    _formatting.FORMATTERS = old_formatters
+
+
+@pytest.mark.parametrize(
+    "formatters,expected_errors",
+    [
+        (
+            [test_utils.MakeAFormatter(), test_utils.MakeBFormatter()],
+            ["Conflicting formatters"],
+        ),
+        (
+            [test_utils.MakeBFormatter(), test_utils.AddBFormatter()],
+            ["not able to make stable changes"],
+        ),
+        (
+            [
+                test_utils.MakeBFormatter(),
+                test_utils.AddBFormatter(),
+                test_utils.MakeAFormatter(),
+            ],
+            ["Conflicting formatters:", "Diff too intricate to compute"],
+        ),
+    ],
+)
+def test_conflicting_formatters(
+    formatters: list[Formatter],
+    expected_errors: list[str],
+    tmp_path: Path,
+) -> None:
+    """Tests that conflicting formatters raise an error."""
+    tmp_file = tmp_path / "test.py"
+    with open(tmp_file, "w", encoding="utf-8") as f:
+        content = [
+            '"""AAA AA AAA"""',
+        ]
+        f.writelines(content)
+
+    with patched_run(formatters) as run:
+        with pytest.raises(UnstableResultError) as err:
+            run([str(tmp_file)])
+
+    for expect_err in expected_errors:
+        assert expect_err in str(err.value), str(err.value)
--- a/pydocstringformatter/_testutils/__init__.py
+++ b/pydocstringformatter/_testutils/__init__.py
@@ -10,6 +10,7 @@ import pytest
 from pydocstringformatter import run_docstring_formatter
 from pydocstringformatter._formatting import Formatter
 from pydocstringformatter._testutils.example_formatters import (
+    AddBFormatter,
     MakeAFormatter,
     MakeBFormatter,
 )
@@ -83,4 +84,4 @@ Temp file is '{self.file_to_format}'
         return None
 
 
-__all__ = ["FormatterAsserter", "MakeAFormatter", "MakeBFormatter"]
+__all__ = ["FormatterAsserter", "MakeAFormatter", "MakeBFormatter", "AddBFormatter"]
--- a/pydocstringformatter/_testutils/example_formatters.py
+++ b/pydocstringformatter/_testutils/example_formatters.py
@@ -27,3 +27,16 @@ class MakeBFormatter(Formatter):
         token_dict = tokeninfo._asdict()
         token_dict["string"] = token_dict["string"].replace("A", "B")
         return type(tokeninfo)(**token_dict)
+
+
+class AddBFormatter(Formatter):
+    """A formatter that adds Bs."""
+
+    name = "add-b-formatter"
+    style = ["default"]
+
+    def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+        """Add a B to the end of the string."""
+        token_dict = tokeninfo._asdict()
+        token_dict["string"] += "B"
+        return type(tokeninfo)(**token_dict)
--- a/pydocstringformatter/_utils/__init__.py
+++ b/pydocstringformatter/_utils/__init__.py
@@ -7,6 +7,7 @@ from pydocstringformatter._utils.exceptions import (
 from pydocstringformatter._utils.file_diference import compare_formatters, generate_diff
 from pydocstringformatter._utils.find_docstrings import is_docstring
 from pydocstringformatter._utils.find_python_file import find_python_files
+from pydocstringformatter._utils.issue_template import create_gh_issue_template
 from pydocstringformatter._utils.output import print_to_console, sys_exit
 
 __all__ = [
@@ -18,6 +19,7 @@ __all__ = [
     "PydocstringFormatterError",
     "TomlParsingError",
     "UnstableResultError",
+    "create_gh_issue_template",
     "print_to_console",
     "sys_exit",
 ]
--- a/pydocstringformatter/_utils/__init__.py
+++ b/pydocstringformatter/_utils/__init__.py
@@ -2,6 +2,7 @@ from pydocstringformatter._utils.exceptions import (
     ParsingError,
     PydocstringFormatterError,
     TomlParsingError,
+    UnstableResultError,
 )
 from pydocstringformatter._utils.file_diference import compare_formatters, generate_diff
 from pydocstringformatter._utils.find_docstrings import is_docstring
@@ -16,6 +17,7 @@ __all__ = [
     "ParsingError",
     "PydocstringFormatterError",
     "TomlParsingError",
+    "UnstableResultError",
     "print_to_console",
     "sys_exit",
 ]
--- a/pydocstringformatter/_utils/exceptions.py
+++ b/pydocstringformatter/_utils/exceptions.py
@@ -12,3 +12,7 @@ class UnrecognizedOption(PydocstringFormatterError):
 
 class TomlParsingError(PydocstringFormatterError):
     """Raised when there are errors with the parsing of the toml file."""
+
+
+class UnstableResultError(PydocstringFormatterError):
+    """Raised when the result of the formatting is unstable."""


.gitignore
```diff
@@ -2,3 +2,7 @@
 pydocstringformatter.egg-info
 dist/
 docs/_build
+
```
jspaezp, line 5: My bad ... will remove this changes ...

pydocstringformatter/run.py
```diff
@@ -98,21 +84,125 @@ def format_file(self, filename: Path) -> bool:
...
+                enabled[formatter.name] = formatter
+
+        self._enabled_formatters = enabled
+        return enabled
+
+    def apply_formatters(
+        self, token: tokenize.TokenInfo, allowed_formatters: list[str] | None = None
+    ) -> tuple[tokenize.TokenInfo, list[str]]:
+        """Applies formatters to a token.
+
```
jspaezp, line 185: probably i should improve this description to include the fact that it keeps track of what formatters actually change the token

jspaezp:
Looking at the conflicting formatters ....

NumpydocSectionSpacingFormatter collides with ClosingQuotesFormatter

NumpydocSectionSpacingFormatter
https://github.com/jspaezp/pydocstringformatter/blob/858c1dabe6880e7367f692674bdc731eae398fe1/pydocstringformatter/_formatting/formatters_numpydoc.py#L104
generates:
```python
'"""\n    A docstring\n    """'
```

ClosingQuotesFormatter
https://github.com/jspaezp/pydocstringformatter/blob/858c1dabe6880e7367f692674bdc731eae398fe1/pydocstringformatter/_formatting/formatters_default.py#L119
generates:
```python
'"""A docstring"""'
```

Looking at references

Numpydoc differentiates between short summary and extended summary.
https://numpydoc.readthedocs.io/en/latest/format.html#short-summary

where I think the correct out would be ...

```python
# CURRENT
'"""\n    A docstring\n    """'

# Expected
'"""A docstring\n\n    """'
```

Which for sure would be incompatible....

Solution 1: Make NumpydocSectionSpacingFormatter not modify anything if there is only one section.
Solution 2: Add a Formatter.conflcting atttribute (also entails fixing the tests so that one of the args is disabled)
    - Op1: and check for conflicting formatters when calling _Run, warn about the behavior and disable all conflciting formatters. 
    - Op2: same but error out instead of warning (also entails changing the defaults)

Extras:
1. Improve documentation for closing-quotes to make the behavior more explicit.

```
  --beginning-quotes, --no-beginning-quotes
                        Activate or deactivate beginning-quotes: Fix the position of the opening quotes. Styles: default. (default: True)
  --closing-quotes, --no-closing-quotes
                        Activate or deactivate closing-quotes: Fix the position of the closing quotes. Styles: default. (default: True)
```

jspaezp:
`tests/data/format/linewrap_summary/function_docstrings.args`
breaks as well ...

(expectation left, actual right)
image

and adds two newlines...

pydocstringformatter/run.py
```diff
@@ -98,26 +84,148 @@ def format_file(self, filename: Path) -> bool:
                         file=sys.stderr,
                     )
                 with open(filename, "w", encoding="utf-8", newline=newlines) as file:
-                    file.write(tokenize.untokenize(changed_tokens))
+                    file.write(tokenize.untokenize(formatted_tokens))
```
jspaezp, line 87: I renamed this because "changed_tokens" made it sound like it was only the tokens that changed, and not all the tokens after being passed though the formatting

pydocstringformatter/run.py
```diff
@@ -98,26 +84,148 @@ def format_file(self, filename: Path) -> bool:
...
                 sys.stdout.write(
                     _utils.generate_diff(
                         tokenize.untokenize(tokens),
-                        tokenize.untokenize(changed_tokens),
+                        tokenize.untokenize(formatted_tokens),
                         filename_str,
                     )
                 )
 
         return is_changed
```
jspaezp, line 101: I split the following function into the file and the token handling (it also resolved a "too many branches" problem in pylint)

jspaezp:
@DanielNoord I think I am at the point where some design decisions have to be made.
LMK what you think (no rush)

pydocstringformatter/_formatting/_utils.py
```diff
@@ -8,3 +12,18 @@ def is_rst_title(summary: str) -> bool:
     if not (last_line := summary.splitlines()[-1].lstrip()):
         return False
     return last_line.count(last_line[0]) == len(last_line)
+
+
+def _compare_formatters(
```
DanielNoord, line 17: Since this module name is private we can make this function itself "non-private".

```suggestion
def compare_formatters(
```

pydocstringformatter/_formatting/_utils.py
```diff
@@ -8,3 +12,18 @@ def is_rst_title(summary: str) -> bool:
...
+    token: tokenize.TokenInfo,
+    formatter_1: Formatter,
+    formatter_2: Formatter,
+    title_extra: str = "",
+) -> str:
+    """Modifies a token with two formatters and returns the difference."""
+    out_t1 = formatter_1.treat_token(token)
+    out_t2 = formatter_2.treat_token(out_t1)
+
+    title = f"{formatter_1.name} vs {formatter_2.name} {title_extra}"
```
DanielNoord, line 27:
```suggestion
    title = f"{formatter_1.name} vs {formatter_2.name}"
    if title_extra:
        title += " " + title_extra
```

pydocstringformatter/run.py
```diff
@@ -98,26 +84,148 @@ def format_file(self, filename: Path) -> bool:
...
+            new_tokeninfo = tokeninfo
 
-        return is_changed
+            if _utils.is_docstring(new_tokeninfo, tokens[index - 1]):
+                new_tokeninfo, changers, stable = self.apply_formatters_twice(
+                    new_tokeninfo
+                )
+                is_changed = is_changed or bool(changers)
+
+                # Try again to get a stable result, skipping the formatters that
```
DanielNoord, line 133: I think this might be problematic.

Let's say we have:
```python
"""My summary
is two lines.
"""
```

This formatter that puts a period at the end of the summary won't do anything in this case.
However, if after the first run the summary splitter creates:

```python
"""My summary

is two lines.
"""
```
Then the period added should be called again.

So, I don't think we can skip formatters based on results in a previous round.

pydocstringformatter/run.py
DanielNoord: Just to help reviewing this, if you want you could separate that change into a separate PR so that we could approve and merge that quickly and make the diff of this PR a little smaller.

pydocstringformatter/run.py
```diff
@@ -98,26 +84,148 @@ def format_file(self, filename: Path) -> bool:
...
+                # didn't change the tokens in former passes
+                new_tokeninfo, changers, stable = self.apply_formatters_twice(
+                    new_tokeninfo, changers
+                )
+
+                if not stable:
+                    if len(changers) > 1:
+                        msg = "Conflicting formatters: " + ", ".join(changers)
+                    else:
+                        msg = "Formatter: " + changers[0]
```
DanielNoord, line 143: I like that we also give a the name of the formatter that is "broken". Good idea!

pydocstringformatter/run.py
```diff
@@ -98,26 +84,148 @@ def format_file(self, filename: Path) -> bool:
...
+                        msg += f"\n{diff}"
+
+                    raise RuntimeError(msg)
+
+            formatted_tokens.append(new_tokeninfo)
+
+        return formatted_tokens, is_changed
+
+    @property
+    def enabled_formatters(self) -> dict[str, _formatting.Formatter]:
```
DanielNoord, line 164: Is there a reason why we don't just generate this in the `__init__` of the class? Then we can remove the `property` and private attribute. We have access to `config` at some point in the `__init__`.

tests/data/config/non_valid_toml_boolopt/pyproject.toml
```diff
@@ -3,4 +3,5 @@ write = false
 style = ["numpydoc"]
 strip-whitespaces = true
 split-summary-body = false
+closing-quotes = false
```
DanielNoord, line 6: I'd prefer updating the example code.

tests/data/format/linewrap_summary/function_docstrings.args
```diff
@@ -1 +1,2 @@
 --linewrap-full-docstring
+--max-summary-lines=2
```
DanielNoord, line 2: Same as above, where possible I think the code should be updated instead of the arguments. That makes the test more easily "interpretable". You can assume that all default values are being used instead of having to check every `.args` file to see which specific configuration each test is using.

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,98 @@
...
+import tokenize
+from typing import Generator
+
+import pytest
+from _pytest.fixtures import SubRequest
+
+from pydocstringformatter._formatting.base import Formatter
+
+
+class MakeAFormatter(Formatter):
```
DanielNoord, line 11: 👍

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def treat_token(self, tokeninfo: tokenize.TokenInfo) -> tokenize.TokenInfo:
+        """Add a B to the end of the string."""
+        token_dict = tokeninfo._asdict()
+        token_dict["string"] += "B"
+        return type(tokeninfo)(**token_dict)
+
+
+@pytest.fixture(scope="function")
+def patched_run(request: SubRequest) -> Generator[type, None, None]:
```
DanielNoord, line 51:
```suggestion
def patched_run(request: SubRequest) -> Generator[type[_Run], None, None]:
```
right?

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,98 @@
...
+@pytest.mark.parametrize(
+    "patched_run,expect_error",
+    [
+        ([MakeAFormatter(), MakeBFormatter()], "Conflicting formatters"),
+        ([MakeBFormatter(), AddBFormatter()], "not able to make stable changes"),
+    ],
+    indirect=["patched_run"],
+)
+def test_conflicting_formatters(
+    patched_run: type,  # pylint: disable=redefined-outer-name
```
DanielNoord, line 82:
```suggestion
    patched_run: type[_Run],  # pylint: disable=redefined-outer-name
```

tests/test_formatting_utils.py
```diff
@@ -0,0 +1,49 @@
+import tokenize
+
+from pydocstringformatter._formatting._utils import _compare_formatters
+from pydocstringformatter._formatting.base import Formatter
+
+
+class MakeAFormatter(Formatter):
```
DanielNoord, line 7: We might want to create a `test/_test_utils.py` file so we don't have to duplicate this code.

DanielNoord (changes requested):
Thanks a lot @jspaezp there are many good changes in here.

I did a first review. Let me know what you think!

tests/test_conflicting_formatters.py
jspaezp: Indeed! To be completely honest I did not know how to annotate "the class itself and not an instance of it". Thanks

tests/test_formatting_utils.py
jspaezp: good idea

pydocstringformatter/run.py
jspaezp: https://github.com/DanielNoord/pydocstringformatter/pull/174

pydocstringformatter/run.py
jspaezp: I gave this a lot of thought and this is why every pass is "twice" (hence apply_formatters_twice).

For example if our formatters are ...
[AtoB, BtoC, CtoA]
for "B"
first pass: [B>C, C>A]
second pass: [A>B,B>C,C>A]

Therefore all the formatters will be flagged as changers in the "first double pass". So all of them will be used again in L145.

The tradeoff here is catching some extremely complicated multi-formatter conflict that would only arise after 3 full passes vs speed.

pydocstringformatter/run.py
jspaezp: Well I did this this way thinking about https://github.com/DanielNoord/pydocstringformatter/pull/29 where eventually the setting up and the running would need to be decoupled.

The private attribute is just an explicit cache for the function.
LMK if you want me to move it to `__init__` ... maybe we could do a "middle of the road approach", where I rename this to `get_enabled_formatters` and set its result to `self.enabled_formatters` during `__init__`

tests/data/config/non_valid_toml_boolopt/pyproject.toml
jspaezp: As it is right now it is not possible, those two are not compatible. (numpydoc and closing quotes, because numpydoc enables NumpydocSectionSpacingFormatter)

I think out loud about it here:
https://github.com/DanielNoord/pydocstringformatter/pull/173#issuecomment-1249930115

tests/data/format/linewrap_summary/function_docstrings.args
jspaezp: this one can be done

pydocstringformatter/run.py
DanielNoord: Take:
[DtoB, AtoB, BtoC, CtoA]
for "D"
first pass: [D>B,B>C, C>A]
second pass: [A>B,B>C,C>A]

So the `DtoB` won't be called again in the third case. Imo, that time save is negligible in comparison to the complexity this adds. `pydocstringformatter` is (currently) quite fast anyways. Especially when there are no changes to write which will be the case for most projects after they have done their initial setup. I'd prefer just running all conflicts in all 3 runs and having that slight bit of potential delay.
It might also be hard to debug issues if not all 3 of the runs are the same. You'd have to figure out which formatters are turned on when to do a full analysis which might be annoying.

pydocstringformatter/run.py
DanielNoord: Yeah I think the last suggestion makes sense. It's a bit too long to put in the `__init__` but I think we should use a helper function/method to set it during the `__init__`.

For #29 we could do:
1. load config
2. load additional formatters
3. load enabled formatters

DanielNoord:
> Looking at the conflicting formatters ....
> 
> NumpydocSectionSpacingFormatter collides with ClosingQuotesFormatter

```python
# CURRENT
'"""\n    A docstring\n    """'

# Expected
'"""A docstring"""'
```


I think the Expected should be as above. This is what PEP256 mandates for single line docstrings. `Numpy` refers to the PEP in its introduction so I think they will accept that this is indeed the correct "output" for these examples.

Would you wan to create an issue/PR to fix this?


> ## I think this is a feature and not a bug ...
> why: because it reconciles "max_summary_lines" and "max_line_length" ... since the summary first gets broken due to their max-length and then the space gets added because the max_summary_lines is 1 !!!

Agreed. Perhaps we should switch the order around? So the wrapping is done before the splitting?

pydocstringformatter/run.py
jspaezp: alright I see your point.
Will remove that filter in the "second double pass"

pydocstringformatter/run.py
jspaezp: OK, I will get that in the next commit

pydocstringformatter/run.py
jspaezp: Done

jspaezp:
Regarding https://github.com/DanielNoord/pydocstringformatter/pull/173#issuecomment-1250238259, i think we can reference this for another issue.

DanielNoord:
> Regarding [#173 (comment)](https://github.com/DanielNoord/pydocstringformatter/pull/173#issuecomment-1250238259), i think we can reference this for another issue.

Agreed!

pydocstringformatter/run.py
```diff
@@ -113,11 +100,129 @@ def format_file(self, filename: Path) -> bool:
...
+
+                        msg += f"\n{diff}"
+
+                    raise RuntimeError(msg)
+
+            formatted_tokens.append(new_tokeninfo)
+
+        return formatted_tokens, is_changed
+
+    def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
```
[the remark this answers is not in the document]
jspaezp: It still applies the style by filtering out the formatters whose style "is not active" (in the configuration)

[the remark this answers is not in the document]
DanielNoord: Hm, but we were doing that already though right? This seems to add some methods and attribute for something that we currently do in a for loop of 3/4 lines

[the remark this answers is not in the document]
jspaezp: Well to be more explicit, it replaces L72-L79 in the parent (loop inside 'format file'). For this method.

In my opinion it abstracts away the "getting the enabled formatters, provided the current configuration" and makes sure it gets called only once during initialization (instead of for every file).

So i think It trades off some "lines" for conceptual clarity on the logic of the formatter.
If this is something you are not interested in, we could move it back to the main function. LMK if you would like me to do that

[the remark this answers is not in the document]
DanielNoord: You're right! Doing this only once makes sense.

That said, perhaps we should extract it into its own PR? It should be much easier to review and get it in.

Do you know why the tests are failing?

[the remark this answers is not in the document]
jspaezp: 1. Wohooo
2. Moved it to PR: https://github.com/DanielNoord/pydocstringformatter/pull/178
3. I am not 100% sure, but I think the `type(CLASSS)` synmtax was added in 3.9, pre 3.9 I THINK `typing.Type[CLASS]` should work (not sure why the linters did not pick that up) ...

pydocstringformatter/run.py
```diff
@@ -123,11 +113,117 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
 
         return enabled
 
-    def format_files(self, filepaths: list[Path]) -> bool:
```
jspaezp, line 126: Note, Format files was moved to the end of the class, diffs here are misleading

pydocstringformatter/run.py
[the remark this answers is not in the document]
jspaezp: Yup that was it

tests/conftest.py
```diff
@@ -9,3 +13,39 @@ def test_file(tmpdir: LocalPath) -> str:
     with open(filename, "w", encoding="utf-8") as file:
         file.write('"""A multi-line\ndocstring."""')
     return str(filename)
+
+
+def make_a_formatter() -> Formatter:
```
DanielNoord, line 18: I'd prefer to just import from `_test_utils`. I tend to forget which functions and fixtures I defined in `conftest` and I don't like having to refer back to `conftest` while browsing my tests.

tests/data/config/non_valid_toml_boolopt/pyproject.toml
DanielNoord: We probably need to fix #177 before we can merge this PR.

pydocstringformatter/run.py
```diff
@@ -123,11 +113,117 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+
+        tokens: List of tokens to format.
+        filename: Name of the file the tokens are from.
+
+        Returns:
+            A tuple containing [1] the formatted tokens in a list
+            and a boolean indicating if the tokens were changed.
+
+        Raises:
+            RuntimeError:
```
DanielNoord, line 129: I'd prefer to make a `UnstableResultError` which inherits from `PydocstringFormatterError`.
See:

pydocstringformatter/run.py
```diff
@@ -123,11 +113,117 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+            if allowed_formatters and formatter_name not in allowed_formatters:
+                continue
+
+            if (new_token := formatter.treat_token(token)) != token:
+                changers.append(formatter_name)
+                token = new_token
+
+        return token, changers
+
+    def format_files(self, filepaths: list[Path]) -> bool:
```
DanielNoord, line 226: Just to get the diff stable we might also want to do this refactor in a separate PR.

pydocstringformatter/run.py
```diff
@@ -123,11 +113,117 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
 
-        return is_changed
+            if _utils.is_docstring(new_tokeninfo, tokens[index - 1]):
+                new_tokeninfo, changers, stable = self.apply_formatters_twice(
+                    new_tokeninfo
+                )
+                is_changed = is_changed or bool(changers)
+
+                # Try again to get a stable result, skipping the formatters that
+                # didn't change the tokens in former passes
```
DanielNoord, line 149: This comment needs to be updated.

pydocstringformatter/run.py
```diff
@@ -123,11 +113,117 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+                new_tokeninfo, changers, stable = self.apply_formatters_twice(
```
DanielNoord, line 150: We should only apply once here.

tests/conftest.py
jspaezp: actually this is my bad, I did not intend to stage this file

pydocstringformatter/run.py
```diff
@@ -123,11 +113,117 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+                    new_tokeninfo,
+                )
+
+                if not stable:
+                    if len(changers) > 1:
+                        msg = "Conflicting formatters: " + ", ".join(changers)
+                    else:
+                        msg = "Formatter: " + changers[0]
+                        msg += "Is not able to make stable changes on:\n"
```
DanielNoord, line 159: Could you add something about "Please open an issue on Github with the following info ....".

Perhaps create a `make_github_issue_template` function in `utils` so that we can create a neat looking multi-line `f-string` that has all relevant information.

pydocstringformatter/run.py
```diff
@@ -123,11 +113,117 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+            List of formatters to apply.
+            If None, all enabled formatters are applied.
+
+        Returns:
+            A tuple containing [1] the formatted token and [2] a list
+            of formatters that changed the token.
+        """
+        changers = []
+        for formatter_name, formatter in self.enabled_formatters.items():
+            if allowed_formatters and formatter_name not in allowed_formatters:
```
DanielNoord, line 217: Are we doing something with `allowed_formatters` anymore? I think it is left over from when we filtered formatters.

pydocstringformatter/run.py
jspaezp: Done

pydocstringformatter/run.py
jspaezp: It is done internally if no changes happen in the first pass. in "apply_formatters_twice". The name might be slightly missleading but the name "apply_formatters_once_or_twice_if_unstable" was not that appealing

pydocstringformatter/run.py
jspaezp: Done

pydocstringformatter/run.py
jspaezp: removed it

tests/data/config/non_valid_toml_boolopt/pyproject.toml
jspaezp: i think this is mergeable and we can have 177 be part of the milestones for the next release

pydocstringformatter/_utils/issue_template.py
```diff
@@ -0,0 +1,75 @@
...
+        formatters: The formatters that caused the issue.
+        filename: The filename of the file that caused the issue.
+    """
+    formatter_names = list(formatters)
+    msg = ""
+    if len(formatter_names) > 2:
+        msg = f"""
+Conflicting formatters: {", ".join(formatters)}
+"""
+        diff = None
```
DanielNoord, line 25: Why is `diff` `None` here? Can't we generate a `diff`?

pydocstringformatter/run.py
```diff
@@ -112,30 +113,101 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+                    new_tokeninfo
+                )
+                is_changed = is_changed or bool(changers)
+
+                # Run formatters again to check if the result is stable
+                new_tokeninfo, changers, stable = self.apply_formatters_twice(
+                    new_tokeninfo,
+                )
+
+                if not stable:
```
DanielNoord, line 149: Can't this be simplified to just check `changers` again? We could remove the return of `stable` and just depend on whether `changers` is `True` in the third run.

pydocstringformatter/run.py
```diff
@@ -112,30 +113,101 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+
+        Returns:
+            A tuple containing:
+            [1] the formatted token and
+            [2] a list of formatters that changed the token.
+            [3] a boolean indicating if the token was in the second pass.
+        """
+        stable = False
+        token, changers = self.apply_formatters(token)
+        if changers:
```
DanielNoord, line 180: If on the 3rd run there are `changers` won't we start a 4th run here?

pydocstringformatter/run.py
```diff
@@ -112,30 +113,101 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
+
+    def apply_formatters(
+        self, token: tokenize.TokenInfo
+    ) -> tuple[tokenize.TokenInfo, list[str]]:
+        """Applies formatters to a token and keeps track of what changes it.
+
+        token: Token to apply formatters to
+        allowed_formatters:
+            List of formatters to apply.
+            If None, all enabled formatters are applied.
```
DanielNoord, line 196:
```suggestion
```

tests/data/format/linewrap_summary/function_docstrings.py.out
```diff
@@ -1,13 +1,15 @@
 def func():
```
DanielNoord, line 1: This test update can only be done with the `two-pass` enabled. So this can't be extracted to another PR currently.

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,72 @@
+import tempfile
+from typing import Generator, List, Type
```
DanielNoord, line 2:
```suggestion
from __future__ import annotations

from collections.abc import Generator

import tempfile
```

This allows you to use `list[]` and `type[]`.

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,72 @@
...
+
+import pytest
+from _pytest.fixtures import SubRequest
+
+from pydocstringformatter import _testutils as test_utils
+from pydocstringformatter._utils.exceptions import UnstableResultError
```
DanielNoord, line 8:
```suggestion
from pydocstringformatter._utils import UnstableResultError
```

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,72 @@
...
+from pydocstringformatter.run import _Run
+
+
+@pytest.fixture(scope="function")
+def patched_run(request: SubRequest) -> Generator[Type[_Run], None, None]:
+    """Patches the _Run class to use the MakeAFormatter and MakeBFormatter."""
+    # This monkeypatches the formatters to make them conflict.
+    # and the only formatters applied
+    # pylint: disable=wrong-import-position, import-outside-toplevel
+    from pydocstringformatter import _formatting  # noqa: E402
```
DanielNoord, line 18: I think this can be moved outside of the `fixture`.

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,72 @@
...
+        ),
+    ],
+    indirect=["patched_run"],
+)
+def test_conflicting_formatters(
+    patched_run: Type[_Run],  # pylint: disable=redefined-outer-name
+    expect_errors: List[str],
+) -> None:
+    """Tests that conflicting formatters raise an error."""
+
```
DanielNoord, line 59:
```suggestion
```

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,72 @@
...
+    with tempfile.TemporaryDirectory() as tmp:
```
DanielNoord, line 60: You can use the `tmp_path` `fixture` from `pytest`. We use it in other tests as well.

DanielNoord (changes requested):
I extracted some changes and moved some stuff around to follow the structure of other utilities and functions.

This is nearing completion I think. I'll have a look somewhere today if I can fix those conflicting formatters myself so that the mitigations against those don't need to be part of this PR.

tests/data/config/non_valid_toml_boolopt/pyproject.toml
DanielNoord: I'm removing these settings as I have merged #181 and they are now unnecessary 🎉

pydocstringformatter/_utils/issue_template.py
jspaezp: I set it to None and it stays that way if there are more than 2 problematic formatters. Otherwise it gets over-written in L49

tests/test_conflicting_formatters.py
jspaezp: I was not sure if having it import outside would lead to the "global" _Run be patched as well as the patched one, will give it a try and move it if it works

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,72 @@
...
+    old_formatters = _formatting.FORMATTERS
+    _formatting.FORMATTERS = request.param
+
+    def cleanup() -> None:
+        """Returns the formatters to their original state."""
+        _formatting.FORMATTERS = old_formatters
+
+    request.addfinalizer(cleanup)
+
+    yield _Run
```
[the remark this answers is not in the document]
jspaezp: I believe this makes sure the cleanup runs even if there is an error (will also try it out and move it if it works)

pydocstringformatter/_utils/issue_template.py
DanielNoord: Can't we generate a diff with more than 2 formatters?

jspaezp: well ... Im not sure ...
1. If you are talking about making a multi-way diff, I dont believe we can ... I do not know how ...
2. I think we could do several sequential 2-formatter comparissons, but I think the output would be massive
3. since it shows the input token and the problematic formatters I don't think we need to show also a diff when many formatters show problems.

DanielNoord: Okay, but let's put something there like:
"Diff too intricate to compute for this amount of formatters".

pydocstringformatter/run.py
DanielNoord: Yeah, but my question is regarding your previous comment where you said that this functions makes sure not to apply formatters a fourth time. I think this line makes it so it is not true.
If run 1 has changers, run 2 will be triggered by this check.
If run 2 has changers, run 3 will be triggered in the calling function.
If run 3 has changers, then this check starts a run 4. I think we want to avoid that.

pydocstringformatter/run.py
jspaezp: OK

pydocstringformatter/run.py
jspaezp: OK

pydocstringformatter/run.py
jspaezp: I think I follow changed it so it only applies once after the first double change

tests/test_conflicting_formatters.py
jspaezp: OK, I was wrong, modified it

tests/test_conflicting_formatters.py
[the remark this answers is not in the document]
jspaezp: seems to work

tests/test_conflicting_formatters.py
jspaezp: Done

pydocstringformatter/_utils/issue_template.py
```diff
@@ -0,0 +1,76 @@
...
+        filename: The filename of the file that caused the issue.
+    """
+    formatter_names = list(formatters)
+    msg = ""
+    if len(formatter_names) > 2:
+        msg = f"""
+Conflicting formatters: {", ".join(formatters)}
+"""
+        diff = "Diff too intricate to compute for this"
+        diff += f" number of formatters ({len(formatter_names)})"
```
DanielNoord, line 26:
```suggestion
        diff = f"Diff too intricate to compute for {len(formatter_names)} formatters."
```

pydocstringformatter/run.py
```diff
@@ -112,30 +113,92 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
 
+    def apply_formatters(
+        self, token: tokenize.TokenInfo
+    ) -> tuple[tokenize.TokenInfo, list[str]]:
+        """Apply the formatters twice to a token.
+
+        Also tracks which formatters changed the token.
+
+        Returns:
+            A tuple containing:
```
DanielNoord, line 171: The `tuple` only has two elements.

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,71 @@
...
+from _pytest.fixtures import SubRequest
+
+from pydocstringformatter import _formatting
+from pydocstringformatter import _testutils as test_utils
+from pydocstringformatter._utils import UnstableResultError
+from pydocstringformatter.run import _Run
+
+
+@pytest.fixture(scope="function")
+def patched_run(request: SubRequest) -> Generator[type[_Run], None, None]:
```
DanielNoord, line 16: `request` is now unused right?

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,71 @@
...
+
+    old_formatters = _formatting.FORMATTERS
+    _formatting.FORMATTERS = request.param
+
+    def cleanup() -> None:
+        """Returns the formatters to their original state."""
+        _formatting.FORMATTERS = old_formatters
+
+    yield _Run
+    cleanup()
```
DanielNoord, line 29: Let's just do this instead of using an intermediate function.

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.mark.parametrize(
+    "patched_run,expect_errors",
```
DanielNoord, line 33:
```suggestion
    "patched_formatters,expected_errors",
```

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,71 @@
...
+        ),
+    ],
+    indirect=["patched_run"],
+)
+def test_conflicting_formatters(
+    patched_run: type[_Run],  # pylint: disable=redefined-outer-name
+    expect_errors: list[str],
+    tmp_path: Path,
+) -> None:
+    """Tests that conflicting formatters raise an error."""
```
DanielNoord, line 59: This is quite an elegant test 😄

DanielNoord (changes requested):

pydocstringformatter/run.py
jspaezp: mb

tests/test_conflicting_formatters.py
jspaezp: done

tests/test_conflicting_formatters.py
jspaezp: still used in L22 to use it as a reusable fixture

tests/test_conflicting_formatters.py
jspaezp: thanks :), took em a while to come up with it ...

tests/test_conflicting_formatters.py
DanielNoord: Oh my bad, but can't that be a normal param? Something like:
```suggestion
def patched_run(formatters: list[str]) -> Generator[type[_Run], None, None]:
```

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,71 @@
...
+                test_utils.AddBFormatter(),
+                test_utils.MakeAFormatter(),
+            ],
+            ["Conflicting formatters:", "Diff too intricate to compute"],
+        ),
+    ],
+    indirect=["patched_run"],
+)
+def test_conflicting_formatters(
+    patched_run: type[_Run],  # pylint: disable=redefined-outer-name
```
DanielNoord, line 55: I just noticed, the typing seems incorrect here. Shouldn't this be `list[Formatters]`?

tests/test_conflicting_formatters.py
jspaezp: well .... if we moved it from being a fixture to a context manager ... and then just used it directly in the test

DanielNoord: I think even normal fixtures can return `Generators`? We use them in `pylint` as well. And we pass normal parameters then as well.

tests/test_conflicting_formatters.py
jspaezp: nope, it is actually the _Run class

tests/test_conflicting_formatters.py
jspaezp: Ok, reading the pytest documentation made clear how to ... but it is awful to read ...
Changed it to a simple contextlib manager, let me know what you think (same result, much more readable)

pydocstringformatter/run.py
```diff
@@ -112,30 +113,92 @@ def get_enabled_formatters(self) -> dict[str, _formatting.Formatter]:
...
     ) -> tuple[list[tokenize.TokenInfo], bool]:
-        """Format a list of tokens."""
+        """Format a list of tokens.
+
+        tokens: List of tokens to format.
+        filename: Name of the file the tokens are from.
+
+        Returns:
+            A tuple containing [1] the formatted tokens in a list
+            and a boolean indicating if the tokens were changed.
```
DanielNoord, line 125:
```suggestion
            and [2] a boolean indicating if the tokens were changed.
```

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,69 @@
...
+from pydocstringformatter._utils import UnstableResultError
+from pydocstringformatter.run import _Run
+
+
+@contextmanager
+def patched_run(formatters: list[Formatter]) -> Generator[type[_Run], None, None]:
+    """Patches the _Run class to use only the provided formatters."""
+    # This monkeypatches the formatters to make them conflict.
+    # and the only formatters applied
+
```
DanielNoord, line 21:
```suggestion
```

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,69 @@
...
+                test_utils.MakeBFormatter(),
+                test_utils.AddBFormatter(),
+                test_utils.MakeAFormatter(),
+            ],
+            ["Conflicting formatters:", "Diff too intricate to compute"],
+        ),
+    ],
+)
+def test_conflicting_formatters(
+    formatters: list[Formatter],  # pylint: disable=redefined-outer-name
```
DanielNoord, line 52:
```suggestion
    formatters: list[Formatter],
```

DanielNoord (changes requested):
Looks much better as a context manager!

tests/test_conflicting_formatters.py
```diff
@@ -0,0 +1,66 @@
...
+from pydocstringformatter._formatting import Formatter
+from pydocstringformatter._utils import UnstableResultError
+from pydocstringformatter.run import _Run
+
+
+@contextmanager
+def patched_run(formatters: list[Formatter]) -> Generator[type[_Run], None, None]:
+    """Patches the _Run class to use only the provided formatters."""
+    old_formatters = _formatting.FORMATTERS
+    try:
```
DanielNoord, line 20: Why are we in a `try` here?

jspaezp: To be honest, its to be safe. it makes sure the cleanup happens if there are un-handled errors while yielding the patched run (which I'm not sure it it would affect all other tests).

https://docs.python.org/3/library/contextlib.html

DanielNoord: Hm, for now I'd say let's remove this. If any error occurred while patching `formatters` (so while we run this test) the test (and test suite) will fail anyway. It's not necessary to guarantee a successful test suite run as if we raise we will fail anyway.

I did learn something new today though, that's always nice 😄

jspaezp: Right on!
Alright, it works after removing it in [862356d](https://github.com/DanielNoord/pydocstringformatter/pull/173/commits/862356d89f3d7c39f078839963c3313c0ffb421b)

DanielNoord (commented):
I'm going to give this one final review tomorrow when I'm "fresh", but I think we ironed out all the kinks 😄 🎉

DanielNoord (approved):
@jspaezp I rebased and reordered the commit history to make it a bit cleaner. I also did some changes to capitalisation and changed `changers` to be a `set[str]` instead of `list[str]`. If you agree with the current state of the PR I think we can merge 😄

jspaezp:
those seem like good changes!

Issue #177, opened 2022-09-20
NumpydocSectionSpacingFormatter collides with ClosingQuotesFormatter


jspaezp:
It was discussed: #173 
(check the whole conversation if you want more context)

1. NumpydocSectionSpacingFormatter collides with ClosingQuotesFormatter
https://github.com/DanielNoord/pydocstringformatter/pull/173#issuecomment-1249930115

> Looking at the conflicting formatters ....
> 
> NumpydocSectionSpacingFormatter collides with ClosingQuotesFormatter
> 
> NumpydocSectionSpacingFormatter https://github.com/jspaezp/pydocstringformatter/blob/858c1dabe6880e7367f692674bdc731eae398fe1/pydocstringformatter/_formatting/formatters_numpydoc.py#L104 generates:
> 
> ```python
> '"""\n    A docstring\n    """'
> ```
> 
> ClosingQuotesFormatter https://github.com/jspaezp/pydocstringformatter/blob/858c1dabe6880e7367f692674bdc731eae398fe1/pydocstringformatter/_formatting/formatters_default.py#L119 generates:
> 
> ```python
> '"""A docstring"""'
> ```
> 
> Looking at references
> 
> Numpydoc differentiates between short summary and extended summary. https://numpydoc.readthedocs.io/en/latest/format.html#short-summary
> 
> where I think the correct out would be ...
> 
> ```python
> # CURRENT
> '"""\n    A docstring\n    """'
> 
> # Expected
> '"""A docstring\n\n    """'
> ```
> 
> Which for sure would be incompatible....
> 
>     * Solution 1: Make `NumpydocSectionSpacingFormatter` not modify anything if there is only one section.
> 
>     * Solution 2: Make `NumpydocSectionSpacingFormatter` by making it call the `ClosingQuotesFormatter` on its current output.
> 
>     * Solution 2: Add a Formatter.conflcting atttribute (also entails fixing the tests so that one of the args is disabled)
>       
>       * Op1: and check for conflicting formatters when calling _Run, warn about the behavior and disable all conflciting formatters.
>       * Op2: same but error out instead of warning (also entails changing the defaults)
> 
> 
> Extras:
> 
>     1. Improve documentation for closing-quotes to make the behavior more explicit.
> 
> 
> ```
>   --beginning-quotes, --no-beginning-quotes
>                         Activate or deactivate beginning-quotes: Fix the position of the opening quotes. Styles: default. (default: True)
[...]

PR #181, opened 2022-09-22
Don't add a new line in one-line Numpy docstrings

DanielNoord:
Closes https://github.com/DanielNoord/pydocstringformatter/issues/177


DanielNoord:
Tagging @jspaezp as you might be interested in the fact that I found a fix 😄

PR #278, opened 2023-05-15
Add numpydoc section name checker

DanielNoord:
Follow up to https://github.com/DanielNoord/pydocstringformatter/pull/248.

I think the idea of that PR should still be pursued and I don't want to lose the branch. I haven't figured out a good way to do it though. How can we display that `pydocstringformatter` can't format the code in a nice way..


DanielNoord:
Closing as I won't be pursuing this for now.

